PR #18, opened 2017-05-02
Fix out of bounds array error caused by shrink_segment

bernie-simon:
Drizzle computes the overlap between each line of the input image and the output image. Part of the computation is shrink_segment, which reduces the length of the input image line if there are NaN values along the edge of the image. That code mistakenly assumed that the size of the input image and pixel map were the same, leading to the out of bounds error when the pixel map was larger. The code in shrink_segment has been modified to intialize the line length from the input image and not the pixel map.

Some logging code has been added to cdrizzleutil.h and various functions to make catching errors easier in the future. The code is controlled by a C macro, LOGGING, and is usually turned off. There is a new global variable, driz_log_handle, defined in cdrizzleapi, initiallized to NULL.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,7 +35,7 @@ env:
         - PIP_DEPENDENCIES=''
         - EVENT_TYPE='pull_request push'
 
-	# For this package-template, we include examples of Cython modules,
+    # For this package-template, we include examples of Cython modules,
         # so Cython is required for testing. If your package does not include
         # Cython code, you can set CONDA_DEPENDENCIES=''
         - CONDA_DEPENDENCIES='Cython'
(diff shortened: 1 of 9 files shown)


jdavies-st:
Just to follow up on this.  This fixed 2 problems I was seeing:

- `Segmentation fault: 11` when calling the `drizzle.tdriz` code where happening inconsistently
- Rows of NaNs and large random values were being mapped to the output frame 1 pixel beyond the extent of each input image.

Excellent fix.

mcara:
> That code mistakenly assumed that the size of the input image and pixel map were the same

@jdavies-st Do you happen to know why this is not a valid assumption?

mcara:
Also see https://github.com/spacetelescope/drizzle/blob/8c7452d2493c37b219fc21eceef43bdfd476e271/src/cdrizzleapi.c#L221-L225

It looks like `image size == pixel map` size is enforced in the code.

jdavies-st:
Hi @mcara, what's the context here?  It looks like this is an old PR that fixed problems with WCS objects that produced NaNs, i.e. the gwcs.WCS instances for the JWST spectroscopic modes that have NaN mappings outside of the bounding box.  In these cases, the specific row of the mapping gets shrunk (ends trimmed off) if there are NaNs at the end.  So both the array and the mapping needed to match.

PR #20, opened 2017-05-17
Add pixmap dimension check to drizzle

bernie-simon:
The drizzle code expects that the dimensions of the input image and pixmap file are the same. The blotting code expects the dimensions of the output image and pixmap file are the same. Prior to this change, the C code would throw a segmentation violation if the pixmap did not meet this condition. This change adds a check and throws an explicit error if the condition is not met. If your pixmap is larger than the input/output image, it can be trimmmed to match the dimensions. If it is smaller, resize it to have the correct dimensions and fill the undefined pixels with NaN and the function shrink_segment will see that only the corresponding section of the input image is used in the
drizzle. Some diagnostic prints used in debugging C errors are also modified in this change. They are turned off in production. This new test may explicitly break existing code, but it was going to fail
anyway with a harder to debug segmentation violation.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -96,7 +96,7 @@ matrix:
         # time.
 
         - os: linux
-          env: PYTHON_VERSION=2.7 NUMPY_VERSION=1.7
+          env: PYTHON_VERSION=2.7 NUMPY_VERSION=1.10
         - os: linux
           env: PYTHON_VERSION=3.3 NUMPY_VERSION=1.8
         - os: linux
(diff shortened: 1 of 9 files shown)


mcara:
I think "adding a pixmap dimension check" is a great idea. However, I think expanding/shrinking pixel map is a dangerous thing.

bernie-simon:
It's not the pixel map that changes size. Each line in the input image is checked to see if there is a valid value in the pixmap array and non-zero value in the weights array. If either check fails, that line in the input image is "shrunk" and only the valid pixels are use in the drizzle.

PR #32, opened 2018-11-08
Remove use of deprecated astropy.WCS _naxis1 & _naxis2

jamienoss:
This is a propagation of https://github.com/astropy/astropy/pull/7973

This would tether Drizzle to Astropy >=3.1.

Signed-off-by: James Noss <jnoss@stsci.edu>

### DIFF
--- a/drizzle/doblot.py
+++ b/drizzle/doblot.py
@@ -69,7 +69,7 @@ def doblot(source, source_wcs, blot_wcs, exptime, coeffs = True,
         Was used when input to output mapping was computed
         internally. Is no longer used and only here for backwards compatibility.
     """
-    _outsci = np.zeros(blot_wcs.pixel_shape[::-1],dtype=np.float32)
+    _outsci = np.zeros(blot_wcs.pixel_shape[::-1], dtype=np.float32)
 
     # compute the undistorted 'natural' plate scale 
     wcslin = blot_wcs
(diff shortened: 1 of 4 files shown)


pllim:
@mcara and @nden , I feel like this needs to be reviewed sooner than later, if `jwst` uses this code.

jdavies-st:
Do we not have any CI for this package?

`jwst` doesn't currently use the affected code in the diff, as we don't currently use the FITS WCS wrapper for `drizzle`.  We have our own wrapper for `gwcs`.

But it would be good to eventually get this reviewed and merged.

pllim:
Theoretically, there is Travis CI set up, so something must have broken. But even the tests that ran were all failing... 🤷‍♀️ 

https://travis-ci.org/spacetelescope/drizzle/branches

drizzle/doblot.py
```diff
@@ -69,7 +69,7 @@ def doblot(source, source_wcs, blot_wcs, exptime, coeffs = True,
         Was used when input to output mapping was computed
         internally. Is no longer used and only here for backwards compatibility.
     """
-    _outsci = np.zeros((blot_wcs._naxis2,blot_wcs._naxis1),dtype=np.float32)
+    _outsci = np.zeros(blot_wcs.pixel_shape[::-1],dtype=np.float32)
```
jdavies-st: @jamienoss, any chance you could make a small change to this PR, such as putting a space before `dtype` here?  That will kick of the CI tests and maybe we can get this merged.  Thanks!  😄 

In case you haven't been following the saga, the Travis CI tests on this repo have been broken for, um, a while.  Now they're back.

jdavies-st: Thanks!  Tests pass.  Merging.

PR #44, opened 2019-03-28
Stop packaging the test data

### LINKED ISSUE #42: Remove test files from distributed package
Currently the test FITS files are included in the MANIFEST, which means that the `pip install` size is 47MB.  It should be less than 1MB.  Also, the test files don't need to be 11MB in size.
```
rey> pip install drizzle --no-cache-dir
Collecting drizzle
  Downloading https://files.pythonhosted.org/packages/63/a8/d73561940986e75b3874e35d9353d1be5734d62daefbff912d53850bd98d/drizzle-1.12.tar.gz (47.0MB)
Requirement already satisfied: astropy in /Users/jdavies/miniconda3/envs/test/lib/python3.7/site-packages (from drizzle) (3.1.2)
Requirement already satisfied: numpy>=1.13.0 in /Users/jdavies/miniconda3/envs/test/lib/python3.7/site-packages (from astropy->drizzle) (1.16.2)
Installing collected packages: drizzle
  Running setup.py install for drizzle ... done
Successfully installed drizzle-1.12
```
Maybe move the tests out of the distributed package too.

jhunkeler:
Knocks off a few hundred MBs

Fixes #42.

Closes #46.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -9,8 +9,11 @@ show-response = 1
 
 [tool:pytest]
 minversion = 3.0
-norecursedirs = build docs/_build astropy_helpers
+norecursedirs = .eggs build docs/_build astropy_helpers
 doctest_plus = enabled
+
+[aliases]
+test=pytest
 
 [ah_bootstrap]
 auto_use = True
--- a/setup.cfg
+++ b/setup.cfg
@@ -7,6 +7,28 @@ all_files = 1
 upload-dir = docs/_build/html
 show-response = 1
 
+[flake8]
+max-line-length = 160
+exclude =
+    build,
+    .eggs,
+    .svn,
+    CVS,
+    .bzr,
+    .hg,
+    .git,
+    __pycache__
+select =
+    F,
+    E101,
+    E111,
+    E112,
+    E113,
+    E401,
+    E402,
+    E711,
+    E722
+
 [tool:pytest]
 minversion = 3.0
 norecursedirs = .eggs build docs/_build astropy_helpers
--- a/setup.py
+++ b/setup.py
@@ -107,6 +107,10 @@ TESTS_REQUIRE = [
     'pytest-runner'
 ]
 
+DOCS_REQUIRE = [
+    'sphinx'
+]
+
 setup(name=PACKAGENAME,
       version=VERSION,
       description=DESCRIPTION,
@@ -115,7 +119,8 @@ setup(name=PACKAGENAME,
       install_requires=['Cython', 'astropy'],
       tests_require=TESTS_REQUIRE,
       extras_require={
-          'test': TESTS_REQUIRE
+          'test': TESTS_REQUIRE,
+          'docs': DOCS_REQUIRE
       },
       author=AUTHOR,
       author_email=AUTHOR_EMAIL,
--- a/setup.py
+++ b/setup.py
@@ -39,7 +39,7 @@ LONG_DESCRIPTION = package.__doc__
 builtins._ASTROPY_PACKAGE_NAME_ = PACKAGENAME
 
 # VERSION should be PEP386 compatible (http://www.python.org/dev/peps/pep-0386)
-VERSION = '1.13.dev'
+VERSION = '1.13'
 
 # Indicates if this version is a release version
 RELEASE = 'dev' not in VERSION
@@ -51,6 +51,11 @@ if not RELEASE:
 # invoking any other functionality from distutils since it can potentially
 # modify distutils' behavior.
 cmdclassd = register_commands(PACKAGENAME, VERSION, RELEASE)
+
+# Obliterate astropy's test hook
+if cmdclassd.get('test'):
+    del cmdclassd['test']
+
 
 # Adjust the compiler in case the default on this platform is to use a
 # broken one.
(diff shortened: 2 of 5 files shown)


jhunkeler:
So maybe it isn't too surprising but

## This works
```bash
$ pytest drizzle
```

## This doesn't
```bash
$ python setup.py test
```

pllim:
```
E       FileNotFoundError: [Errno 2] No such file or directory: '/tmp/drizzle-test-g4uotdyb/lib/python3.6/site-packages/drizzle/tests/data/input1.fits'
```
If you stop defining the data as package data, then you also have to rethink how a test is supposed to use the data, because the current `DATA_DIR` lookup no longer works.

pllim:
What I originally had in my mind was:

1. Move `tests` outside of `drizzle` to the root directory, like what your attempt with `jwst`.
2. Then test data can still be "packaged" as test data but the whole `tests` tree will not be distributed as part of MANIFEST.

Does this make sense?

pllim:
p.s. I am actually not sure if the test runner that comes with `astropy-helpers` can support a test suite outside of package dir. 😬

jdavies-st:
It can't as far as I can tell.  Tried it already.

pllim:
> It can't as far as I can tell.

In that case, if you want to move the `tests` outside of package dir, then you need to ditch the test runner and patch whatever needs patching so that you can run this directly using `pytest tests` or something similar, which is probably more work than you anticipated?

jhunkeler:
> Then test data can still be "packaged" as test data but the whole tests tree will not be distributed as part of MANIFEST.

Just so we're clear on why this has to happen. We can't use `astropy_hellpers`' data injection here because `wheel` (and other `bdist_*` commands) consumes whatever is defined as `package_data`. `MANIFEST.in` manipulates what's included or excluded from a `sdist` tarball/zip, and `package_data` manipulates which files (usually non-python related) get packaged during compilation.

I'll just obliterate the `"test"` command class. `pytest` doesn't appear to have an issue running these outside of astropy's testing framework.

pllim:
Re: obliterate

Before that happens, we need to stare deep into its soul and ask ourselves whether it really needs all the fancy plugins like `pytest-openfiles`, `pytest-remotedata`, and `pytest-doctestplus`. If not, then things will be easy. If yes, then you have to make sure they still work the way you refactor stuff.

jdavies-st:
I doesn't need any of those fancy plugins.  Documentation is minimal, there's no remote data.  I tried ditching the test runner in #44, but there's a bunch of hidden state stuff it does that needs more sorting out.  Ugh.  Not a lot of time to do that right now, but I can get back on it if Joe's solution here hits a dead end.

The real question is, does this package even need `astropy_helpers` at all anyway? Seems like a lot of restrictive baggage, at least from the testing side.

It would be nice to have `python setup.py test` work too, at least in the cloned repo.

pllim:
> does this package even need `astropy_helpers`

Strictly speaking, no package ever *needs* the helpers. But gutting it does require some work, as you might have to refactor your setup files.

jdavies-st:
🎉 🍾

jdavies-st:
Wow!  `python setup.py test` works now.  What sort of magical incantations did you do.  Can you explain so that I can understand?

setup.py
```diff
@@ -39,7 +39,7 @@
 builtins._ASTROPY_PACKAGE_NAME_ = PACKAGENAME
 
 # VERSION should be PEP386 compatible (http://www.python.org/dev/peps/pep-0386)
-VERSION = '1.13.dev'
```
pllim: Why remove the dev tag?

pllim:
> What sort of magical incantations

I noticed several things:

* Not-so-subtle obliteration of Astropy test runner.
* Aliasing `test` command to `pytest` executable.
* Getting rid of `ci-helpers` completely and `pip install` all dependencies.

It is too bad you don't straight up remove the submodule. :smirk: :clap:

setup.py
jdavies-st: I think Joe wants to release this.

setup.cfg
```diff
@@ -7,11 +7,36 @@ all_files = 1
...
+    E722
+
 [tool:pytest]
 minversion = 3.0
-norecursedirs = build docs/_build astropy_helpers
+norecursedirs = .eggs build docs/_build astropy_helpers
 doctest_plus = enabled
 
+[aliases]
+test=pytest
```
jdavies-st: Is this safe?

pllim: `#livingdangerously`

setup.py
pllim: I don't know how I feel about sneaking in a release tag in this PR, but I am not the one doing the release. Just as long as release going to happen immediately after merge *and* "dev" is added back after the release.

setup.cfg
jdavies-st: Ah, I see.  Because it uses the pytest-runner, it works.  Cool.  I withdraw my concern.

jdavies-st: The trick is you have to do:
```
pip install -e .[test]
python setup.py test
```
Without `pytest-runner` package, the error message from `python setup.py test` is a bit opaque.

pllim: At this point, do we even need to use `setup.py test`? Isn't calling `pytest` directly less confusing?

jdavies-st: The current recommended best practice seems to be moving away from extra commands such as `test` and `build_sphinx` implemented in setup.py and instead moving to just running pytest at the top level.  This allows setup.py to almost become a one-liner and setup.cfg to contain all declarative setup of the package.

Anyway, happy to do this in stages.  =)

jdavies-st:
Shall we merge and do a release @jhunkeler?

PR #47, opened 2019-06-19
No cruft

jhunkeler:
- Attempt to decouple from `astropy_helpers`
- Add `pyproject.toml`


jdavies-st:
I suspect that once this is merged we can tag with `1.13.1` and do a release to pypi?  Is there a good reason not to build wheels for this package?

PR #54, opened 2020-10-30
Remove appveyor.yml

### LINKED ISSUE #35: TST: Clean up Travis CI matrix
I think the current Travis CI matrix was copied blindly from an old Astropy project template. It is not optimized for testing this package and needs some serious clean-up.

One example: Cron jobs are defined but it is not actually set up to run in the CI service.

### LINKED ISSUE #53: MNT: Stop using ci-helpers in appveyor.yml
To whom it may concern,

If you are using https://github.com/astropy/ci-helpers in your appveyor.yml , please know that the Astropy project has dropped active development/support for Appveyor CI. If it still works, good for you, because we did not remove the relevant files (yet). But if it ever stops working, we have no plans to fix anything for Appveyor CI. Please consider using native Windows support other CI, e.g., Travis CI (see https://docs.travis-ci.com/user/reference/windows/). We apologize for any inconvenience caused.

If this issue is opened in error or irrelevant to you, feel free to close. Thank you.

xref astropy/ci-helpers#464

jdavies-st:
Resolves #53 
Resolves #35


.travis.yml
```diff
@@ -1,57 +1,56 @@
...
-      - libopenblas-base
-      - libopenblas-dev
-
-env:
-  global:
-    - NUMPY_VERSION=1.16
-    - DEV="--upgrade cython git+https://github.com/numpy/numpy git+https://github.com/astropy/astropy"
-
-matrix:
+python: 3.8.5
```
pllim, line 3: Is pinning bugfix version necessary?
```suggestion
python: 3.8
```

.travis.yml
```diff
@@ -1,57 +1,56 @@
...
+        - "pip install -e .[docs]"
+      script:
+        - sphinx-build docs docs/_build
+      addons:
+        apt:
+          packages:
+            - libopenblas-base
+            - libopenblas-dev
+
+    - name: Dev dependencies
```
pllim, line 28: Might want to do this one as cron job but to now allow to fail, up to you.

In my experience, allowing something to fail cause people not to notice until it is too late, especially if the developer is often busy with something else.

pllim (approved):
Just minor comments. LGTM as long as the CI passes.

.travis.yml
jdavies-st: I think generally, if you specify only `3.8` it will use `3.8.0`.  At least that has been my experience with Travis.  Also, I recall in the past that it interprets `3.8` as an integer, while it interprets `3.8.5` as a string, and that has different effects.  Maybe if I do `"3.8"` it will get the latest Python 3.8.x?

.travis.yml
jdavies-st: Agree.  I will remove the allowed failure, but keep it on a per-PR for now.  I don't want to have to fiddle with cron settings in the Travis setup right now.

.travis.yml
pllim: Hmm... `astropy` not using quotes, but then again it also uses `tox`. 🤷 

https://github.com/astropy/astropy/blob/3bae6fa88a9dbc07f1aab696cb47decca1ee21cd/.travis.yml#L76

jdavies-st: https://travis-ci.com/github/astropy/astropy/jobs/417355775#L237

pllim: Hmmm... But it is not 3.7.0. 🤷 

You might want to open astropy issue then. One more reason to move away from Travis CI? 😬

jdavies-st: Maybe there's some default patch that Travis picks if you  just specify the minor release.  Regardless, I don't think it's a problem to specify a later patch.  For `3.8` it installs `3.8.1`:

https://travis-ci.com/github/astropy/astropy/jobs/417355772#L247

Anyway, I'll keep it as-is, as it doesn't hurt.

pllim (approved):
Feel free to merge when CI passes.

p.s. You don't want to test against `os: osx` and `os: windows`? 😸

PR #56, opened 2021-01-05
Adding github actions for continuous integration.

kmacdonald-stsci:
Adding github actions.

### DIFF
--- a/.github/workflows/drizzle_ci.yml
+++ b/.github/workflows/drizzle_ci.yml
@@ -1,0 +1,81 @@
+name: Drizzle Package Continuous Integration
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  build:
+    name: ${{ matrix.name }}
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        include:
+          - name: Python 3.9 Test
+            pyver: 3.9
+            dependencies: 
+
+          - name: Old Dependencies
+            pyver: 3.6.10
+            dependencies: |
+              python -m pip install numpy~=1.16.0
+
+          - name: Development Build
+            pyver: 3.8.5
+            dependencies: |
+              python -m pip install -i https://pypi.anaconda.org/scipy-wheels-nightly/simple --pre -I numpy
+              python -m pip install git+https://github.com/astropy/astropy
+    steps:
+    - uses: actions/checkout@v2
+    - name: ${{ matrix.name }}
+      uses: actions/setup-python@v2
+      with:
+        python-version: ${{ matrix.pyver }}
+    - name: Install Dependencies
+      run: |
+        python -m pip install --upgrade pip
+        python -m pip install -e .[test]
+        ${{ matrix.dependencies }}
+    - name: Test with Pytest
+      run: |
+        pytest
+
+  doc_build:
+    name: Document Build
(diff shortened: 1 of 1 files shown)


.github/workflows/drizzle_ci.yml
```diff
@@ -0,0 +1,81 @@
+name: Drizzle Package Continuous Integration
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
```
jdavies-st, line 7: I think we want it to run on all pull requests, not just those against master, and we also want it to run when we push tags.  So something like:

```yaml
on:
  push:
    branches:
      - master
    tags:
      - "*"
  pull_request:
```

.github/workflows/drizzle_ci.yml
```diff
@@ -0,0 +1,81 @@
...
+    - uses: actions/checkout@v2
+    - name: Set up Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.9
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        python -m pip install flake8
+    - name: Build documents with sphinx
```
jdavies-st, line 79: Update this name description to be linting-related.  "Run flake8".  Also, I don't think this is doing anything, as it is looking for `asdf`, not anything in `drizzle`.

.github/workflows/drizzle_ci.yml
```diff
@@ -0,0 +1,81 @@
...
+            pyver: 3.9
+            dependencies: 
+
+          - name: Old Dependencies
+            pyver: 3.6.10
+            dependencies: |
+              python -m pip install numpy~=1.16.0
+
+          - name: Development Build
+            pyver: 3.8.5
```
jdavies-st, line 26: This dev build should probably run on Python 3.9 too.

.github/workflows/drizzle_ci.yml
```diff
@@ -0,0 +1,81 @@
+name: Drizzle Package Continuous Integration
```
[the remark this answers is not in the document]
kmacdonald-stsci: This gets printed only once at the top of the page.  Each test is named, so the test names get printed for each test.

.github/workflows/drizzle_ci.yml
kmacdonald-stsci: I changed the name and changed the test to run on `drizzle`, instead of `asdf`.

.github/workflows/drizzle_ci.yml
kmacdonald-stsci: I made these changes.

.github/workflows/drizzle_ci.yml
kmacdonald-stsci: I updated to Python 3.9.

.github/workflows/drizzle_ci.yml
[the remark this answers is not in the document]
jdavies-st: This text gets printed for every check.  I.e.:

Screen Shot 2021-01-07 at 2 33 02 PM

Preferable would just be "CI" as in `jwst`:

Screen Shot 2021-01-07 at 2 33 33 PM

As it is easier to read and find the job names.

jdavies-st (approved):
Looks good.

PR #58, opened 2021-01-11
Add coverage to CI

kmacdonald-stsci:
Adding code coverage to github actions.

### DIFF
--- a/.github/workflows/drizzle_ci.yml
+++ b/.github/workflows/drizzle_ci.yml
@@ -99,7 +99,7 @@ jobs:
         python -m pip install -e .[test]
     - name: Test with pytest and code coverage
       run: |
-        pytest --cov-report=xml --cov=. --cov-config=setup.cfg
+        pytest --cov-report=xml --cov=. --cov-config=./drizzle/tests/coveragerc
     - name: Upload coverage to codecoverage
       uses: codecov/codecov-action@v1
       with:
--- a/.github/workflows/drizzle_ci.yml
+++ b/.github/workflows/drizzle_ci.yml
@@ -97,7 +97,6 @@ jobs:
       run: |
         python -m pip install --upgrade pip
         python -m pip install -e .[test]
-        python -m pip install pytest-cov
     - name: Test with pytest and code coverage
       run: |
         pytest --cov-report=xml --cov=. --cov-config=setup.cfg
@@ -105,4 +104,3 @@ jobs:
       uses: codecov/codecov-action@v1
       with:
         file: ./coverage.xml
-        flags: unit
--- a/.github/workflows/drizzle_ci.yml
+++ b/.github/workflows/drizzle_ci.yml
@@ -99,7 +99,7 @@ jobs:
         python -m pip install -e .[test]
     - name: Test with pytest and code coverage
       run: |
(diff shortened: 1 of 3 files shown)


.github/workflows/drizzle_ci.yml
```diff
@@ -82,3 +82,27 @@ jobs:
...
+    - uses: actions/checkout@v2
+    - name: Set up Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.9
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        python -m pip install -e .[test]
+        python -m pip install pytest-cov
```
jdavies-st, line 100: Again, `pytest-cov` should go into the `[test]` extras dependencies in setup.py so it doesn't have to be installed here.

.github/workflows/drizzle_ci.yml
```diff
@@ -82,3 +82,27 @@ jobs:
...
+    - name: Test with pytest and code coverage
+      run: |
+        pytest --cov-report=xml --cov=. --cov-config=setup.cfg
+    - name: Upload coverage to codecoverage
+      uses: codecov/codecov-action@v1
+      with:
+        file: ./coverage.xml
+        flags: unit
```
jdavies-st, line 108: No flags are needed.

.github/workflows/drizzle_ci.yml
```diff
@@ -82,3 +82,27 @@ jobs:
...
+      with:
+        python-version: 3.9
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        python -m pip install -e .[test]
+        python -m pip install pytest-cov
+    - name: Test with pytest and code coverage
+      run: |
+        pytest --cov-report=xml --cov=. --cov-config=setup.cfg
```
jdavies-st, line 103: May want to add some configuration in setup.cfg to make sure coverage is covering what we expect.

jdavies-st (commented):
Would it be possible to fold coverage into the existing tests runs of old and latest dependencies instead of adding another test run independently?

.github/workflows/drizzle_ci.yml
kmacdonald-stsci: I made this change.

.github/workflows/drizzle_ci.yml
kmacdonald-stsci: I removed flags.

.github/workflows/drizzle_ci.yml
kmacdonald-stsci: I changed this to be `--cov-config=./drizzle/tests/coveragerc`.

jdavies-st: Unfortunately that `coveragerc` file isn't doing anything because it looks like Bernie just copied the template there without running `cookiecutter` on it.  I.e. the `{packagename}` stuff is making it not do anything.

jdavies-st: You can see it is not having an effect by clicking through to the Codecov report below and looking at covered files:

https://codecov.io/gh/spacetelescope/drizzle/pull/58/tree?path=drizzle

jdavies-st (approved):
🚀

PR #60, opened 2021-02-22
Use oldest-supported-numpy in pyproject.toml

jdavies-st:
Instead of using the latest version of `numpy` as the build dependency, use the oldest compatible one via the meta package

https://github.com/scipy/oldest-supported-numpy

so that the compiled code can be used with any new version of `numpy` in the runtime environment.

This should resolve the issue discussed in https://github.com/spacetelescope/jwst/issues/5744

For more detail, see https://github.com/pypa/pip/issues/9542

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,10 @@
 [build-system]
-requires = ["setuptools>=42", "wheel", "numpy", "setuptools_scm[toml]>=3.4"]
+requires = [
+    "setuptools>=42",
+    "setuptools_scm[toml]>=3.4",
+    "wheel",
+    "oldest-supported-numpy",
+]
 build-backend = "setuptools.build_meta"
 
 [tool.setuptools_scm]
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -1,5 +1,10 @@
 [build-system]
-requires = ["setuptools>=42", "wheel", "numpy", "setuptools_scm[toml]>=3.4"]
+requires = [
+    "setuptools>=42",
+    "setuptools_scm[toml]>=3.4",
+    "wheel",
+    "oldest-supported-numpy",
```
eslavich, line 6: Well that is a handy dandy feature isn't it

eslavich (approved):
LGTM

pyproject.toml
jdavies-st: We may want to do this for other `spacetelescope` projects that have C code that need `numpy` installed to build/compile.  Maybe we should file issues for the others as well.

PR #66, opened 2021-06-08
Fix memory corruption in the linear blot algorithm

mcara:
Fixes an issue with bilinear interpolator used in the blot function due to which code was trying to access elements outside of the data array resulting in segfault.

This effectively fixes https://github.com/spacetelescope/jwst/issues/6116

This PR also adds a changelog file to this repository though it starts with current issue.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,0 +1,15 @@
+.. _release_notes:
+
+=============
+Release Notes
+=============
+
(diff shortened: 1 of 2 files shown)


mcara:
CC: @nden @jdavies-st @stsci-hack

jdavies-st (approved):
Looks great.  Fixes the bug, cleans up the code.

🚀

PR #67, opened 2021-06-14
Cleanup

jdavies-st:
- PEP8 cleanup and check all the `flake8` codes in CI by default from now on
- Don't rely on tests being run in a particular order.  Use fixtures for 2 tests that depend on a single result.
- Use `tmpdir` for test output instead of home-grown solution
- Remove python 2 cruft


eslavich (approved):
LGTM!

PR #69, opened 2021-06-14
Fix context plane shape bug

### LINKED ISSUE #50: Dimensionality issue when a new plane is added to the context
There is a dimension mismatch error on line 474 of drizzle.py when there are more than 32 input images. We fixed this by replacing this line with 

self.outcon = np.append(self.outcon, [plane], axis=0)

jdavies-st:
Resolves #50


eslavich (approved):
LGTM

mcara (approved):
LGTM

PR #79, opened 2021-12-21
Fix incorrect drizzling when inwht is 0

mcara:
This PR fixes a bug in the C code of the core drizzle algorithm due to which flux from the input image could be dropped onto the output (resampled) image even though those input pixels have zero weights.

The same bug is present in the C code in the `drizzlepac`.

Fixes https://github.com/spacetelescope/jwst/issues/6167

### DIFF
--- a/drizzle/tests/test_drizzle.py
+++ b/drizzle/tests/test_drizzle.py
@@ -257,8 +257,8 @@ def test_zero_input_weight(tmpdir):
     """
     insci = np.ones((200, 400), dtype=np.float32)
     inwht = np.ones((200, 400), dtype=np.float32)
-    inwht[:, 150:153] = 0
-    pixmap = np.rollaxis(np.mgrid[1:201, 1:401][::-1], 0, 3)
+    inwht[:, 150:155] = 0
+    pixmap = np.moveaxis(np.mgrid[1:201, 1:401][::-1], 0, -1)
 
     outsci = np.zeros((210, 410), dtype=np.float32)
     outwht = np.zeros((210, 410), dtype=np.float32)
@@ -278,7 +278,7 @@ def test_zero_input_weight(tmpdir):
         fillstr='INDEF'
     )
 
-    assert np.allclose(np.sum(outsci), 200 * (400 - 3))
+    assert np.allclose(np.sum(outsci), 200 * (400 - 5))
 
 
 def test_square_with_grid(tmpdir):
--- a/drizzle/tests/test_drizzle.py
+++ b/drizzle/tests/test_drizzle.py
@@ -258,27 +258,30 @@ def test_zero_input_weight(tmpdir):
     insci = np.ones((200, 400), dtype=np.float32)
     inwht = np.ones((200, 400), dtype=np.float32)
     inwht[:, 150:155] = 0
+    tflux = np.sum(inwht)
     pixmap = np.moveaxis(np.mgrid[1:201, 1:401][::-1], 0, -1)
 
-    outsci = np.zeros((210, 410), dtype=np.float32)
-    outwht = np.zeros((210, 410), dtype=np.float32)
-    outcon = np.zeros((210, 410), dtype=np.int32)
+    for kernel in ['square', 'point', 'turbo']:
+        # initialize output:
+        outsci = np.zeros((210, 410), dtype=np.float32)
+        outwht = np.zeros((210, 410), dtype=np.float32)
+        outcon = np.zeros((210, 410), dtype=np.int32)
 
-    _vers, nmiss, nskip = cdrizzle.tdriz(
-        insci, inwht, pixmap,
-        outsci, outwht, outcon,
-        uniqid=1,
-        xmin=0, xmax=400,
-        ymin=0, ymax=200,
-        pixfrac=1,
-        kernel='point',
-        in_units='cps',
-        expscale=1,
-        wtscale=1,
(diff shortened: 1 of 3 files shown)


mcara:
@nden @stsci-hack @hbushouse Please review

src/cdrizzlebox.c
```diff
@@ -26,18 +26,21 @@ inline_macro static int
...
-  const double vc_plus_dow = vc + dow;
+  double vc_plus_dow;
   
-  /* Just a simple calculation without logical tests */
-  if (vc == 0.0) {
+  if (dow == 0.0f) return 0;
+  
+  vc_plus_dow = vc + dow;
+  
+  if (vc == 0.0f) {
```
PaulHuwe, line 35: Being C, wouldn't this potentially get tripped up by floating point error?

mcara: No (not for our our purpose). 0.0 is 0.0 exactly. So if some pixel in the output weight array (that is what `vc` really is) has value exactly 0.0 - then all it means that we can skip the fancy (complete) computation done in the `else` block below and just assign `dow` (input pixel weight) to `vc` (output pixel weight). If "somehow" (due to floating errors or "for real")`vc !=0` then the code will proceed with full computations in the `else` block. Nothing dangerous here.

mcara: No (not for our purpose). 0.0 is 0.0 exactly. So if some pixel in the output weight array (that is what `vc` really is) has value exactly 0.0 - then all it means that we can skip the fancy (complete) computation done in the `else` block below and just assign `d` (input pixel) to the output array (`output_data`) without doing weighted sum and assign `vc` (output pixel weight) to the output weight array (`output_counts`) at the end of the code. If "somehow" (due to floating errors or "for real")`vc !=0` then the code will proceed with full computations in the `else` block. Nothing dangerous here.

drizzle/tests/test_drizzle.py
```diff
@@ -251,6 +251,36 @@ def test_square_with_point(tmpdir):
...
 
 
+def test_zero_input_weight(tmpdir):
+    """
+    Test do_driz square kernel with grid
+    """
+    insci = np.ones((200, 400), dtype=np.float32)
+    inwht = np.ones((200, 400), dtype=np.float32)
+    inwht[:, 150:153] = 0
+    pixmap = np.rollaxis(np.mgrid[1:201, 1:401][::-1], 0, 3)
```
PaulHuwe, line 261: rollaxis is deprecated, with moveaxis being the preferred replacement.

src/cdrizzlebox.c
```diff
@@ -26,18 +26,21 @@ inline_macro static int
...
             const float d, const float vc, const float dow) {
 
-  const double vc_plus_dow = vc + dow;
+  double vc_plus_dow;
   
-  /* Just a simple calculation without logical tests */
-  if (vc == 0.0) {
+  if (dow == 0.0f) return 0;
+  
+  vc_plus_dow = vc + dow;
```
PaulHuwe, line 33: If dow is zero, does this function exit and thus never reach lines 54-59 (particularly line 58)? Is it safe to assume that the data will have already been updated for vc (i.e. there is no situation that skipping line 58 here won't be an issue?))?

PaulHuwe (commented):
Have some more questions.

drizzle/tests/test_drizzle.py
mcara: Thanks for pointing this out. Will update the code soon.

src/cdrizzlebox.c
mcara: Good question. The idea in my fix (this PR) is that if a pixel - an _input_ one - has 0 weight (when considering its contribution to an output pixel) and thus it _is not_ contributing to the _output_ pixel at all, then why do we need to update output pixel at all or the total weight (lines 54-59) of that output pixel? Nothing changes when "dropped flux" has 0 weight...

If you look at `cdrizzlebox.c`, you will see that `vc` is simply the current (before "this" update) weight of an output pixel:
```c
vc = get_pixel(p->output_counts, ii, jj);
```

If the flux that we are going to drop onto that output pixel has a non-zero weight, then we should update that weight (and output pixel value). But if the dropped flux is not contributing to the output pixel I do not think its weight should be modified.

If an output fixel has no flux at all in it (because no input pixel contributed to this output pixel) - its weight will be 0 before and after `update_data()` assuming that during this last update dropped flux has also 0 weight (`dow==0`).

mcara: The mistake in the previous logic (IMO) was that it was testing to see whether output image pixel had a contribution from anything before (`if (vc == 0.0)`) and if nothing ever contributed to that output pixel then the old code would drop input flux regardless of its weight simply _by assigning_ this flux to the output pixel - see the `"then"` block after `if (vc == 0.0)`.

mcara: Interestingly, with the old code, when it would drop flux with 0 weight onto an output pixel with pre-existing weight of 0, then `vc_plus_dow` would be `0 + 0` and so output pixel weight assigned on lines 54-59 would still be 0. And so when another input image would be added (drizzled) to the ouput, it would just overwrite previous value in those 0-weight regions.

PaulHuwe (approved):
LGTM

hbushouse (approved):
LGTM

PR #82, opened 2022-08-02
Fix indexing problems in interpolating the pixmap

stsci-hack:
The interpolation routine for the pixmap values in 'cdrizzlemap.c' needed to be revised to properly account for the indices of the values from the pixmap.  Originally, it defined a set of points for interpolation in 1 dimension, but due to how the pixmap values are stored (in a numpy array?),  the values being interpolated over were actually in the other dimension.  This resulted in the pixfrac input parameter being ignored and a sub-pixel shift being introduced to the output array when a non-default orientation was specified for resampling.  

These changes insure that the requested dimension gets used for the interpolation.  The changes were tested using a subset of 4 observations which not only demonstrated the original problem in comparison to HST data, but also confirmed that the problems got fixed.  

This addresses the problem as described in [Jira ticket JP-2702](https://jira.stsci.edu/browse/JP-2702).

### DIFF
--- a/src/cdrizzlemap.c
+++ b/src/cdrizzlemap.c
@@ -278,8 +278,8 @@ interpolation_bounds(
 
         /* Check if the pixel value is NaN */ 
         oob_pixel(pixmap, xy[0], xy[1]); 
-        double pixval = get_pixmap(pixmap, xy[0], xy[1])[idim];
-    
+        double pixval = get_pixmap(pixmap, xy[0], xy[1])[jdim];
+
         /* If not, copy it to output as a good point */
         if (! npy_isnan(pixval)) {
           for (kdim = 0; kdim < 2; ++kdim) {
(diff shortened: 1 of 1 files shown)


src/cdrizzlemap.c
```diff
@@ -328,8 +334,10 @@ interpolate_point(
   int xypix[4][2];
   double partial[4];
   int ipix, jpix, npix, idim;
+  int jdim;
```
mcara, line 337: I would suggest to simply add `jdim` to the previous line like this:

```c
int ipix, jpix, npix, idim, jdim;
```

mcara (changes requested):
My testing of proposed changes shows shifts of the same magnitude as before except shifts (for my particular images) are to the right and down instead of to the right and up as they used to be before (as if there was a reflection _in shifts_ w.r.t. X-axis).

PR #83, opened 2022-08-04
JP-2702: Fix image shifts with square kernel

mcara:
Closes #82
Resolves https://jira.stsci.edu/browse/JP-2702

This is an alternative PR to #82 and fixes the interpolation algorithm used by the square kernel. First, the interpolation itself is now bilinear. Second, the algorithm for finding vertices of the interpolation rectangle (`interpolation_bounds()`) was replaced with a simplified version `interpolation_starting_point()` that does not have the

> Bounce around the starting point until we find four valid points on the line

logic. That is, if one of the vertices has an `NaN` value then interpolation result will also be an `NaN`. I do not think this should cause problems but it should be checked with images that have WCS with bounding boxes and whose WCS produces NaNs outside of bounding boxes.

My initial testing shows that this PR solves the issue. However, as I mentioned above, it should be tested with WCS that can give NaN outside of the bounding box and also it should be tested with various pixfrac ... parameters.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -3,6 +3,10 @@
 
 - Pin astropy min version to 5.0.4. [#81]
 
+- Fix a bug in the interpolation algorithm used by the 'square' kernel that
+  resulted in shifts of the resampled image typically by 0.5 pixels compared
+  to the location indicated by the WCS. [#83]
+
 1.13.4 (2021-12-23)
 ===================
 
--- a/src/cdrizzlemap.c
+++ b/src/cdrizzlemap.c
@@ -19,7 +19,7 @@
  */
 
 void
-initialize_segment(struct segment *self, integer_t x1, integer_t y1, integer_t x2, integer_t y2) { 
+initialize_segment(struct segment *self, integer_t x1, integer_t y1, integer_t x2, integer_t y2) {
   self->point[0][0] = x1;
   self->point[0][1] = y1;
   self->point[1][0] = x2;
@@ -63,7 +63,7 @@ bad_pixel(PyArrayObject *pixmap, int i, int j) {
       return 1;
     }
   }
-  
+
   return 0;
 }
 
@@ -84,9 +84,9 @@ bad_weight(PyArrayObject *weights, int i, int j) {
       return 1;
     } else {
       return 0;
-    } 
+    }
   }
-  
+
   return 0;
 }
 
@@ -108,7 +108,7 @@ shrink_segment(struct segment *self,
 
   imin = self->point[1][0];
   jmin = self->point[1][1];
-  
+
   for (j = self->point[0][1]; j < self->point[1][1]; ++j) {
     for (i = self->point[0][0]; i < self->point[1][0]; ++ i) {
       if (! is_bad_value(array, i, j)) {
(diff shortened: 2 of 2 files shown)


stsci-hack (approved):
Testing of this PR's changes on actual data confirms that it does produce the expected values for the interpolated output pixel positions, regardless of pixfrac.  Processing a mosaic with the square kernel with ```pixfrac = 1.0``` resulted in:

```
[do_kernel_square] dh: 0.500,  xyin: (1471.500, 990.500)
                                      (1472.500, 990.500)
                                      (1472.500, 989.500)
                                      (1471.500, 989.500)
[map_point] XYIN: (1471.500, 990.500) 
[interpolate_point] pixmap[1471, 990] for (1471.500, 990.500): (1386.724, 1112.644)   XYOUT: (1386.996, 1111.991)
[map_point] XYIN: (1472.500, 990.500) 
[interpolate_point] pixmap[1472, 990] for (1472.500, 990.500): (1386.344, 1111.724)   XYOUT: (1386.616, 1111.071)
[map_point] XYIN: (1472.500, 989.500) 
[interpolate_point] pixmap[1472, 989] for (1472.500, 989.500): (1385.419, 1112.110)   XYOUT: (1385.691, 1111.457)
[map_point] XYIN: (1471.500, 989.500) 
[interpolate_point] pixmap[1471, 989] for (1471.500, 989.500): (1385.800, 1113.030)   XYOUT: (1386.072, 1112.377)
```

For the same mosaic with square kernel and ```pixfrac= 0.1```, we get:

```
[do_kernel_square] dh: 0.050,  xyin: (1471.950, 990.050)
                                      (1472.050, 990.050)
                                      (1472.050, 989.950)
                                      (1471.950, 989.950)
 [map_point] XYIN: (1471.950, 990.050) 
[interpolate_point] pixmap[1471, 990] for (1471.950, 990.050): (1386.724, 1112.644)   XYOUT: (1386.409, 1111.750)
[map_point] XYIN: (1472.050, 990.050) 
[interpolate_point] pixmap[1472, 990] for (1472.050, 990.050): (1386.344, 1111.724)   XYOUT: (1386.371, 1111.658)
[map_point] XYIN: (1472.050, 989.950) 
[interpolate_point] pixmap[1472, 989] for (1472.050, 989.950): (1385.419, 1112.110)   XYOUT: (1386.278, 1111.697)
[map_point] XYIN: (1471.950, 989.950) 
[interpolate_point] pixmap[1471, 989] for (1471.950, 989.950): (1385.800, 1113.030)   XYOUT: (1386.317, 1111.789)
```
This confirms that the correct sub-pixel positions are used for the corners of the square kernel for a given value of ```pixfrac```,  
while also interpolating those output (XYOUT) positions correctly.  Visual comparison of the output images also confirms that this PR works as intended.  In short, everything looks good to me.

mcara:
Many many thanks to @stsci-hack not only for reviewing and testing this PR but also for thew original PR #82 and initial investigation of the root-cause of the issue!

hbushouse:
jwst regression tests run last night using this update showed minor differences in the resampled images and spectra for all instrument modes. By default, the automatic jwst processing uses the square kernel and pixfrac=1.0. Would you expect minor differences due to this update, even when default param values are used?

mcara:
@hbushouse Yes, I would expect differences. Quite significant differences as well: not in the pixel values as such but at their location in the image. Previous interpolation algorithm was very bad, IMO. I think it was interpolating as 1D along each axis instead of whole 2D interpolation and it seems like axes where swapped. Due to this bug, the resulting image had a 0.5 pixels error in the WCS (or think WCS was unchanged but photons fell into wrong bins).

So yes, I expect differences in every pixel regardless of the `pixfrac` value. Only 'square' (default) kernel should be affected, I believe.

PR #89, opened 2023-02-08
Fix incorrect identification of lines to be skipped

mcara:
It was reported (help desk ticket INC0186717) that under some circumstances, some parts of input images may be missing in the resampled output images. This occurs when a vertex of the input image is outside of the resampled image. I traced this to be a bug in how algorithm identifies which lines in the input images are completely outside of the output image.

The image below illustrates the issue. Notice missing data in the top-right corner:

missing-image-data

This PR fixes the issue.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,8 +4,17 @@
 Release Notes
 =============
 
-.. 1.13.7 (unreleased)
+.. 1.13.8 (unreleased)
    ===================
+
+
+1.13.7 (2023-02-09)
+===================
+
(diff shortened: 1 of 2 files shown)


perrygreenfield (approved):
Both Mihai and I don't really understand this code, and doing so will take a lot of time. The changes fix the reported problem so for the time being we want to see if it has any downsides. It is worth creating an issue for further study so that either it is better documented or rewritten to be clearer. A truly interesting corner [ahem] case.

Issue #87, opened 2022-08-13
Python test failures with version 1.13.6


olebole:
While trying to package version 1.13.6 for Debian, I observed these test failures:
```
____________________________ test_square_with_point ____________________________

tmpdir = local('/tmp/pytest-of-pbuilder/pytest-0/test_square_with_point0')

    def test_square_with_point(tmpdir):
        """
        Test do_driz square kernel with point
        """
        output = str(tmpdir.join('output_square_point.fits'))
        output_difference = str(tmpdir.join('difference_square_point.txt'))
    
        input_file = os.path.join(DATA_DIR, 'j8bt06nyq_flt.fits')
        output_template = os.path.join(DATA_DIR, 'reference_square_point.fits')
    
        insci = read_image(input_file)
        inwcs = read_wcs(input_file)
        insci = make_point_image(insci, (500, 200), 100.0)
        inwht = np.ones(insci.shape,dtype=insci.dtype)
        output_wcs = read_wcs(output_template)
    
        driz = drizzle.Drizzle(outwcs=output_wcs, wt_scl="")
        driz.add_image(insci, inwcs, inwht=inwht)
    
        if ok:
            driz.write(output_template)
        else:
            driz.write(output)
            template_data = read_image(output_template)
    
            min_diff, med_diff, max_diff = centroid_statistics("square with point",
                                                               output_difference,
                                                               driz.outsci,
                                                               template_data, 20.0, 8)
    
>           assert med_diff < 1.0e-6
E           assert 0.05702659328654338 < 1e-06

drizzle/tests/test_drizzle.py:251: AssertionError
```
The same failure happens with `test_square_with_grid`.

Environment is the "usual" Debian unstable on x86_64:
* Python 3.10.6
* Astropy 5.1
* Numpy 1.21.5

PR #94, opened 2023-03-31
use `ruff` instead of `flake8`

zacharyburnett:
use `ruff` in tox env and apply fixes from `ruff --fix drizzle/`

### DIFF
--- a/drizzle/tests/test_drizzle.py
+++ b/drizzle/tests/test_drizzle.py
@@ -253,7 +253,7 @@ def test_square_with_point(tmpdir):
 
 
 @pytest.mark.parametrize(
-    'kernel', ['square', 'point', 'turbo', 'gaussian', 'lanczos3']
+    'kernel', ['square', 'point', 'turbo', 'gaussian', 'lanczos3'],
 )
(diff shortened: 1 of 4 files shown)


mcara (approved):

mcara (changes requested):
Why are the unit tests failing?

mcara (approved):

zacharyburnett:
> Why are the unit tests failing?

It looks like they haven't been run in 9 months due to lack of activity; perhaps reference files are out of date?

mcara:
I do not see how this PR would make those tests fail. Interesting, it seems like tests didn't run since August of 2022

zacharyburnett:
https://github.com/spacetelescope/drizzle/actions/runs/4578121504/jobs/8084340122?pr=94#step:6:105
```
E           assert 0.05764503023017435 < 1e-06
```
I'm not sure what this is but I assume it could be fixed with updated truths

zacharyburnett:
@mcara it looks like there's an issue for the failing tests: https://github.com/spacetelescope/drizzle/issues/87

PR #97, opened 2023-04-14
drop support for Python 3.8

zacharyburnett:
Numpy is imminently dropping support for Python 3.8

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,7 +1,7 @@
 [project]
 name = "drizzle"
 description = "A package for combining dithered images into a single image"
-requires-python = ">=3.8"
+requires-python = ">=3.9"
 authors = [
     { name = "STScI", email = "help@stsci.edu" },
 ]
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,8 +28,6 @@ jobs:
     with:
       envs: |
         - linux: test-xdist
-          python-version: 3.8
-        - linux: test-xdist
           python-version: 3.9
         - linux: test-xdist
           python-version: 3.10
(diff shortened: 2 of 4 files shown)


pyproject.toml
```diff
@@ -1,7 +1,7 @@
 [project]
 name = "drizzle"
 description = "A package for combining dithered images into a single image"
-requires-python = ">=3.8"
+requires-python = ">=3.9"
```
mcara, line 4: What I am confused about is that this is basically a forced requirement since the code in this package still would run in Python 3.8. What is the point of this restriction? It is not like we have to do anything to continue supporting it in Python 3.8.

mcara:
So what?

pyproject.toml
zacharyburnett: that is true, however code using Python 3.8 will be using an older version of Numpy, so the results will diverge. We definitely can still keep Python 3.8 and test with it, it might just be more maintenance

mcara: > however code using Python 3.8 will be using an older version of Numpy, so the results / API usage may diverge with new Numpy versions. We definitely can still keep Python 3.8 and test with it, it might just be more maintenance

I do not think this should be happening since all (3.8, 3.9, etc) tests are currently passing simultaneously. However, if you feel keeping 3.8 is more maintenance, then let's remove 3.8.

mcara (approved):

zacharyburnett:
@mcara you're right, we can afford to keep 3.8 for now, if the divergence doesn't introduce major issues. I'll close this PR. However, we do need to figure out the test failures in `>=3.9`

pllim:
I accidentally duplicated this work while investigating devdeps failure (#128).

PR #104, opened 2023-06-28
Fix incorrect overlap computations

### LINKED ISSUE #41: Allow the pixmap to be sized to the output frame
Currently, the 2D image size of the input frames determine the pixmap size, and this is always <= to the output frame size.  But in the case where one would prefer to have a small output frame, i.e. to resample a section of a large input frame into a small cutout into the output, this currently fails.

More details in #30.

### LINKED ISSUE #63: "Drizzling" an image into a smaller one does not produce the expected result
I am using Drizzle to go back and forth between an astronomical scene (the large image), and a small detector that observe in in scanning mode (building a mosaic of the large scene though an on-the-fly observing mode, and I have discovered that when the image provided to Drizzle.add_image() is larger than that that has been used to create the Drizzle instance, the result is incorrect (empty, or not filled completely, this seems to has a dependence on the image size ratio).

here is a test that I have performed with a fits file that I have created for testing purposes. The image contains a recognisable structure of stripes.

hdu = fits.open('Test_Mire.fits')
sceneWcs = WCS(hdu[0].header)
data = hdu[0].data
plt.imshow(data)

This shows correctly the stripe pattern I expect. 

Now I create a smaller image by extracting the central part and updating the WCS so that its is physically centred on the same sky position

data2 = data[50:-50,50:-50].copy()
tmpWcs = sceneWcs.deepcopy()
tmpWcs.pixel_shape = [3759,3866]
tmpWcs.wcs.crpix = [1929.5-50,1983-50]

Then I use drizzle to add that small image in an drizzled map that is created with the WCS of the larger image

drizMap = Drizzle(outwcs=sceneWcs)
drizMap.add_image(data2,tmpWcs)
plt.imshow(drizMap.outsci)

This shows the expected striped pattern to the exception that there is a band of empty pixels on all 4 sides of the field, as indeed the image that I have drizzled has a smaller coverage.

Then I do the reverse, use the smaller image WCS to create the drizzled map and add the larger image into it.

drizMap = Drizzle(outwcs=tmpWcs)
drizMap.add_image(data,sceneWcs)
plt.imshow(drizMap.outsci)

this image is empty (full of 0s). And the context (outcon[0,:,:]) image is also empty as if, as far as add_image is concerned, they have no geometrical overlap. 

I am not attaching my test image because it is quite large but I've discovered the problem when I was trying to make this with much smaller images (by extracting first a section of the large image, large enough still so that the detector would fit into it whatever the orientation, but still quite smaller that the image I have used for this test example).

This looks like a bug.

Marc

### LINKED ISSUE #87: Python test failures with version 1.13.6

mcara:
This is the second attempt to fix the issue described in https://github.com/spacetelescope/drizzle/pull/89. That fix was working only sometimes. This PR is a complete re-design of how `drizzle` computes image overlaps and scan lines.

Resolves #41
Resolves #63 
Fixes #87

Regression tests:
https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/841
https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/872 (for 92b4deb)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,8 +4,18 @@
 Release Notes
 =============
 
-.. 1.13.8 (unreleased)
+.. 1.14.1 (unreleased)
    ===================
+
+
+
+1.14.0 (unreleased)
+===================
+
+- Fixed a bug in how drizzle would compute overlaps between input images and
+  the output image. Due to this bug large parts of input image data may be
+  missing in the resampled data when output image size was set by the
+  caller to be smaller than size needed to hold *all* image data. [#104]
 
 
 1.13.7 (2023-02-09)
--- a/src/cdrizzlemap.c
+++ b/src/cdrizzlemap.c
@@ -442,6 +442,42 @@ simplify_polygon(struct polygon *p) {
 }
 
 
+static void
+orient_ccw(struct polygon *p) {
+    int k, m;
+    struct vertex v1, v2, cm;
+
+    if (p->npv < 3) return;
+
+    // center of mass:
+    for (k = 0; k < p->npv; ++k) {
+        cm.x += p->v[k].x;
+        cm.y += p->v[k].y;
+    }
+    cm.x /= p->npv;
+    cm.y /= p->npv;
+
+    // pick first two polygon vertices and subtract center:
+    v1 = p->v[0];
+    v2 = p->v[1];
+    v1.x -= cm.x;
+    v1.y -= cm.y;
+    v2.x -= cm.x;
+    v2.y -= cm.y;
+
+    if (area(v1, v2) >= 0.0) {
+        return;
+    } else {
+        for (k = 0; k < (p->npv / 2); ++k) {
+            v1 = p->v[k];
+            m = p->npv - 1 - k;
+            p->v[k] = p->v[m];
+            p->v[m] = v1;
+        }
+    }
+}
+
+
 int
 intersect_convex_polygons(const struct polygon *p, const struct polygon *q,
                           struct polygon *pq) {
@@ -455,6 +491,9 @@ intersect_convex_polygons(const struct polygon *p, const struct polygon *q,
     if ((p->npv < 3) || (q->npv < 3)) {
         return 1;
     }
+
+    orient_ccw(p);
+    orient_ccw(q);
 
     if (is_poly_contained(p, q)) {
         *pq = *p;
@@ -905,6 +944,7 @@ init_image_scanner(struct driz_param_t* par, struct scanner *s,
     }
     inpq.npv = pq.npv;
     s->overlap_valid = 1;
+    orient_ccw(&inpq);
 
 _setup_scanner:
 
--- a/src/cdrizzlemap.c
+++ b/src/cdrizzlemap.c
@@ -391,7 +391,7 @@ append_vertex(struct polygon *p, struct vertex v) {
     if ((p->npv > 0) && equal_vertices(p->v[0], v, VERTEX_ATOL)) {
         return 1;
     }
-    if (p->npv >= 2 * IMAGE_OUTLINE_NPTS - 1) {
+    if (p->npv >= 2 * IMAGE_OUTLINE_NPTS) {
         return 1;
     }
     p->v[p->npv++] = v;
--- a/src/cdrizzlemap.c
+++ b/src/cdrizzlemap.c
@@ -20,51 +20,6 @@ static const double APPROX_ZERO = 1.0e3 * DBL_MIN;
 static const double MAX_INV_ERR = 0.03;
 
 
-/** ---------------------------------------------------------------------------
- * Test if a pixmap vaue is bad (NaN)
- *
- * pixmap: the pixel mapping between input and output images
- * i:      the index of a pixel within a line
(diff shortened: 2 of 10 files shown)


nden:
Started a regression test here: https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/772/

mcara:
In the last commit I fixed failing unit tests on Windows. The reason for failure was unrelated to this PR: it was the location of the `drizzle.log` file that was invalid on Windows.

mcara:
This PR is ready for review and/or further testing.

CHANGES.rst
```diff
@@ -4,10 +4,20 @@
 Release Notes
 =============
 
-.. 1.13.8 (unreleased)
+.. 1.14.1 (unreleased)
```
hbushouse, line 7: Why are there 2 unreleased sections in the change log?

hbushouse (approved):
Looks OK to me, but most of my knowledge about drizzling comes from watching rain outside my window.

perrygreenfield (commented):
Only minor changes (none of which affect results).

I'm not happy with how kernels are handled or documented, but that is a preexisting situation that should not be handled by this PR

CHANGES.rst
mcara: that is a comment basically, it does not appear in when rendered. the 1.14.0 will get a date later

src/cdrizzlemap.c
```diff
@@ -13,580 +13,1072 @@
...
+}
 
-  initialize_segment(&inlimit, p->xmin, p->ymin, p->xmax, p->ymax);
-  shrink_segment(&inlimit, p->pixmap, &bad_pixel);
 
-  if (inlimit.invalid == 1) {
-      driz_error_set_message(p->error, "no valid pixels on input image");
-      return 1;
+/**
+ * Map a vertex' coordinates from input frame to the output frame.
```
perrygreenfield, line 959: backwards

src/cdrizzlemap.c
```diff
@@ -13,580 +13,1072 @@
...
+            k = max_left;
+            max_left = max_right;
+            max_right = k;
         }
+    } else {
+        max_left = max_right;
+    }
 
-        xybounds->point[ipoint][idim] = b;
+    // Left: start with minimum and move counter-clockwise:
```
perrygreenfield, line 776: I think the handling of separating the left and right edges could use a little more explanation. For example, a naive reading of this comment seems backwards from what I would expect. In that if we are starting with the edge that is at the y minimum shouldn't we go clockwise instead of counterclockwise (and the same issue for the comment on the right). What is going on with this could be better explained.

perrygreenfield (commented):
A couple more comments.

src/cdrizzlemap.c
```diff
@@ -13,580 +13,1072 @@
...
+}
 
-  initialize_segment(&inlimit, p->xmin, p->ymin, p->xmax, p->ymax);
-  shrink_segment(&inlimit, p->pixmap, &bad_pixel);
 
-  if (inlimit.invalid == 1) {
-      driz_error_set_message(p->error, "no valid pixels on input image");
-      return 1;
+/**
+ * Map a vertex' coordinates from input frame to the input frame.
```
perrygreenfield, line 959: This sounds wrong.

perrygreenfield (commented):
I see one remaining mistake in the comments.

perrygreenfield (approved):
LGTM (aside from previous issues related to kernel issues that should not be addressed by this PR)

PR #110, opened 2023-09-05
Replace polygon intersection with SG polygon clipping algorithm

mcara:
This PR replaces the buggy polygon intersection algorithm with a simpler and more robust Sutherland-Hodgman polygon-clipping algorithm. In testing this algorithm correctly computed intersections that were failing with the current algorithm (test case added to the unit tests).

JWST regression test(s):
https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/902 (c1809d5)
https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/903 (74ab304)


nden:
Does the algorithm require open polygons, i.e. the last vertex is not the same as the first one?

mcara:
> Does the algorithm require open polygons, i.e. the last vertex is not the same as the first one?

Yes, this implementation does. That is because if the polygon is closed then one edge will have length 0.

mcara:
Also other parts of the code _assume_ the polygon is "open".

src/cdrizzlemap.c
```diff
@@ -511,162 +475,109 @@ orient_ccw(struct polygon *p) {
...
+
+                t = t / d;
+                // u = u / d;
+
+                vi.x = pv_->x + (pv->x - pv_->x) * t;
+                vi.y = pv_->y + (pv->y - pv_->y) * t;
+                // vi.x = 0.5 * (pv_->x + wv_->x + (pv->x - pv_->x) * t +
+                //               (wv->x - wv_->x) * u);
+                // vi.y = 0.5 * (pv_->y + wv_->y + (pv->y - pv_->y) * t +
+                //               (wv->y - wv_->y) * u);
```
nden, line 551: Are these lines necessary? And a few lines above?

nden (approved):
Coded algorithm looks OK. Tests and documentation are missing, code needs to be cleaned up. Perhaps another PR.

src/cdrizzlemap.c
mcara: These are not necessary but I am torn on using them. Right now we compute the intersection point along just one line. The point computed along the second line _should_ give the same intersection point. So, to reduce computational errors, it seems it would be better to average the two estimates. However, for our purposes the accuracy from one line, I think, should be enough. So, for now I would like to keep this code commented and maybe I should add a comment to explain what it is doing (or what it would do if uncommented) just in case we encounter a situation that might need it later.

mcara:
All checks passed including docs!

mcara:
An additional, final, regtest for `jwst`: https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/912/

PR #98, opened 2023-04-25
docs fixes

zacharyburnett:

PR #111, opened 2023-09-07
Drizzle rtd

nden:
another try based on Zach's #98


mcara:
why is it taking so long for RTD to build docs?

zacharyburnett:
closed by #112

PR #113, opened 2023-09-09
Reformat C code using clang-format

mcara:
Apply uniform style to C code similar to PEP8 rules for Python code. This uses modified Google style. "external" FCT headers have been moved to a separate folder in order to be excluded from style formatting.


mcara:
There are many lines to review here but this is only style change. No functional changes have been made.

src/cdrizzleapi.c
```diff
@@ -23,425 +23,447 @@
...
+    p.kscale = kscale;
+    p.in_units = unit_cps;
+    p.interpolation = interp;
+    p.ef = ef;
+    p.misval = misval;
+    p.sinscl = sinscl;
+    p.pixmap = map;
+    p.error = &error;
+
+    if (driz_error_check(&error, "xmin must be >= 0", p.xmin >= 0)) goto _exit;
```
kmacdonald-stsci, line 377: I think it would be good to always have brackets around conditional code, looping code, etc., as well as that code always being on the next line.  And regardless whether those standards are adopted, I think whatever is adopted should be uniform.  In particular, there are some lines here where the conditional code is on the same line, whereas other lines have the conditional on the next line.

src/cdrizzleapi.c
```diff
@@ -23,425 +23,447 @@
...
+    return;
 }
 
-/** --------------------------------------------------------------------------------------------------
+/** ---------------------------------------------------------------------------
  * Top level function for drizzling, interfaces with python code
  */
 
 static PyObject *
 tdriz(PyObject *obj UNUSED_PARAM, PyObject *args, PyObject *keywords)
```
kmacdonald-stsci, line 53: Long functions should be broken up into smaller ones.  An attempt should be made to keep functions small enough to be able to see the whole function without needing to scroll.  Historically, that's been assumed to be about 35 lines long.

src/cdrizzleapi.c
```diff
@@ -23,425 +23,447 @@
 static PyObject *gl_Error;
 FILE *driz_log_handle = NULL;
 
-/** --------------------------------------------------------------------------------------------------
+/** ---------------------------------------------------------------------------
  * Multiply each pixel in an image by a scale factor
```
kmacdonald-stsci, line 27: Isn't this type of comment used for autogeneration of documentation?  If so, why not use the tags associated with that like 'at' brief, 'at' param, etc?

src/cdrizzleapi.c
```diff
@@ -452,20 +474,22 @@ invert_pixmap_wrap(PyObject *self, PyObject *args)
...
+    xyout_arr =
+        (PyArrayObject *)PyArray_ContiguousFromAny(xyout, NPY_DOUBLE, 1, 1);
     if (!xyout_arr) {
-      return PyErr_Format(gl_Error, "Invalid xyout array.");
+        return PyErr_Format(gl_Error, "Invalid xyout array.");
     }
 
-    pixmap_arr = (PyArrayObject *)PyArray_ContiguousFromAny(pixmap, NPY_DOUBLE, 3, 3);
+    pixmap_arr =
+        (PyArrayObject *)PyArray_ContiguousFromAny(pixmap, NPY_DOUBLE, 3, 3);
```
kmacdonald-stsci, line 490: If another line is needed for a function call, I think it's better to put the function parameters on the next line, rather than entire assignment.

kmacdonald-stsci (approved):
I am a fan of using coding standards ot impose a uniformity and I generally like the changes that are done for this PR.  I left a few comments for other things that may be considered.  We can talk to others about their coding style preferences for C.  If it's just easier to run some formatter, I am fine with that, too.  

My biggest pet peeve, though, is long functions.  Long functions are hard to follow and hard to maintain.  Functions should be able to fit on a screen without having to scroll.  Long functions do many things that have obvious parts to be broken down into smaller functions.  Enforcing smaller functions also eliminates many of the nested conditionals and loopings, which are difficult to follow, that are currently happening in the long functions.

src/cdrizzleapi.c
mcara: I do not think this is part of a formatting style and I do not know what to think about it. I have no issues with functions being longer than 35 lines.

src/cdrizzleapi.c
mcara: I am fine with both approaches. This particular line has a very long condition and I do not like it - I would have used braces and put `goto _exit;` on the next line. However, something like this:
```C
if (x) continue;
```
is fine with me.

**However**, this formatting will not automatically add extra braces for me and the purpose of this style formatting is to automatically format existing code with minimal effort. There will be a lot work done on the `drizzle` code in the future and as we go we can fix many of these things. I did not want to do all of that (including splitting large functions into shorter ones) in this PR.

mcara:
@kmacdonald-stsci This style tries to follow https://peps.python.org/pep-0007/ vague "guide" but not everything is doable (or recommended - I think we have enough freedom to come up with a style that most agree with).

For example, I could not get this to work automatically:

> When you break a long expression at a binary operator, the operator goes at the end of the previous line, and braces should be formatted as shown. E.g.:
```C
if (type->tp_dictoffset != 0 && base->tp_dictoffset == 0 &&
    type->tp_dictoffset == b_size &&
    (size_t)t_size == b_size + sizeof(PyObject *))
{
    return 0; /* "Forgive" adding a __dict__ only */
}
```

The following style:

> Code structure: one space between keywords like if, for and the following left paren; no spaces inside the paren; braces are required everywhere, even where C permits them to be omitted, but do not add them to code you are not otherwise modifying. All new C code requires braces. Braces should be formatted as shown:
```C
if (mro != NULL) {
    ...
}
else {
    ...
}
```
is the Stroustrup's style and it is taken care of in https://github.com/spacetelescope/drizzle/blob/98d6fa9e32599da6e1ac39adc5e2d124bbd82420/.clang-format#L6

But I do not know how to do switching from Stroustrup to Microsoft Allman style only for very long conditionals. It is either one or another and I like Stroustrup much more than Allman. I like Linux style even more than Stroustrup but I Stroustrup is what PEP7 suggests.

I think we do not have to adhere to PEP 7 since we are not writing code for C implementation of Python.

mcara:
Ah, the 4 spaces indent is taken care of in https://github.com/spacetelescope/drizzle/blob/98d6fa9e32599da6e1ac39adc5e2d124bbd82420/.clang-format#L11-L12

Issue #125, opened 2023-11-26
Missing imports in __init__.py


Lasloruhberg:
With the current __init__.py this package does not properly work for python 3.11 (in jupyter lab).
A line stating 

from . import subpackage_name 

is missing for each subpackage.

### Blocked by

* #130

PR #126, opened 2023-11-26
Local import fixes Issue #125

Lasloruhberg:
Amended the __init__.py to import local modules, such that functionality is maintained (in jupyter lab).

### DIFF
--- a/drizzle/__init__.py
+++ b/drizzle/__init__.py
@@ -1,6 +1,11 @@
 """
 A package for combining dithered images into a single image
 """
+from . import drizzle
+from . import dodrizzle
+from . import doblot
+from . import calc_pixmap
+from . import util
 from importlib.metadata import PackageNotFoundError, version
 
 


pllim:
Hello. Can you please clarify what you mean by `such that functionality is maintained (in jupyter lab)`?

pllim (changes requested):
First of all, I need to understand why this is requested.

Secondly, all the affected modules would need to explicitly define `__all__` to control what gets bubbled up here.

pllim:
Thank you for the clarification! As I have said above, to push this forward, each of the module also needs a clean `__all__` defined. However, I am not a maintainer of this package. I was tagged by a colleague, so you might have to wait a little bit until @mcara tells us which functions in those modules should be exposed at top package level API.

> does not like the unused imports

Add a trailing `# noqa` to each line of the new imports would make checker ignore those lines.

> A module that was compiled using NumPy 1.x cannot be run in NumPy 2.0.0.dev0

That is unrelated to the problem you are trying to solve, so you can ignore it.

mcara:
At this moment I do not think we should modify how this works. This package is in active development and it is very likely to be significantly modified. Once that work is completed, this issue can be revisited. I think it would be better to open an issue (to keep track of this) and close this PR.

pllim:
There is already issue #125 , so closing.

@Lasloruhberg , in the meantime, can you please try a different manner of import? The tests here should give you an idea how the imports are currently happening. Thanks for your interest and patience!

Issue #62, opened 2021-02-26
Set up workflow action to build wheels on different platforms for delivery to PyPI


jdavies-st:
See an example here

https://github.com/spacetelescope/synphot_refactor/blob/master/.github/workflows/publish-to-pypi.yml

jdavies-st:
See

https://github.com/pypa/gh-action-pypi-publish

and

https://github.com/pypa/cibuildwheel

jdavies-st:
And

https://github.com/astropy/wheel-forge

jdavies-st:
See examples here

https://github.com/eslavich/drizzle/blob/master/.github/workflows/build.yml

https://github.com/pypa/cibuildwheel/blob/main/examples/github-deploy.yml

And available `runs-on` platforms here:

https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idruns-on

jdavies-st:
The scikit-learn folks build wheels nightly:

https://github.com/scikit-learn/scikit-learn/blob/main/.github/workflows/wheels.yml

pllim:
There are wheels for Linux and OSX now on https://pypi.org/project/drizzle/#files but looks like Windows is omitted. Not sure why.

pllim:
I added wheels for Windows in #128 🤞

PR #128, opened 2023-11-27
Fix devdeps, other updates

pllim:
**Please use "Squash and Merge" button if you decide to merge this!**

I was trying to fix devdeps and went down the rabbit hole of updating everything else to current standards.

Close #97 

Address Windows portion of https://github.com/spacetelescope/drizzle/issues/62


pllim:
Yeah, your manifest prunes out the test data. The build isolation stuff isn't going to work.

pllim:
Sigh, I ran out of time. Still not sure where this came from. Either editable install mode is refusing to pick up numpy-dev during build or there is real incompatibility in your C-extensions.

Feel free to take over if you want. Most of the diff is generic clean-up, so if you want that in a different PR, lemme know.

```
A module that was compiled using NumPy 1.x cannot be run in
NumPy 2.0.0.dev0 as it may crash. To support both 1.x and 2.x
versions of NumPy, modules must be compiled against NumPy 2.0.
```

mcara:
Impressive! Many thanks! One favor, plase: can you, please move `LICENSE` to the root and relete `licenses` directory?

mcara:
and I will think about one more request just to get to 20 commits

pllim:
Maybe @jhunkeler will have ideas if you can get him to review. 😸

mcara (approved):
wonderful!

mcara:
I felt I needed to steal @jhunkeler brains to review this PR.

pllim:
Fun fact: The job without tox is 3x faster.

PR #134, opened 2024-01-19
API re-design: simplify interface, remove I/O, make instrument agnostic

### LINKED ISSUE #125: Missing imports in __init__.py

### LINKED ISSUE #130: Missing __all__ in modules
All the Python modules should have `__all__` defined. It helps with API doc generation and blind `*` imports.

mcara:
Closes #144
Closes #133
Closes #130
Closes #125

~WARNING: This first commit is the first draft and it is definitely not ready for merging and unit tests have not been updated. Also, documentation is incomplete.~

This PR quite radically changes the API of this package with intention to make this package independent of how data are stored (FITS files, asdf, etc.) and the instrument used (JWST, HST, etc.) and also to reduce the number of functions and duplication and the number of conflicting parameters (or parameters that are changing meaning in different functions).

If some items on the TODO list (in the code) are implemented such as support for weight map of the input image to be `None`, then these changes would result in performance increase for EXPTIME, EXPSQ weighting schemes (as opposite to IVM). I would argue that we go even further with this in not having two parameters for input weights: `weight_map` and `wht_scale`. I think these should combine into a single `weight` which can be either an array or a scalar. IMO, having two parameters is an unnecessary complication.

BUG FIX: exposure time was undefined when `in_units` were not `cps`. That is, `drizzle` was producing random outputs when input image units were, let's say `'counts'`.

BUG FIX: `tdriz` signature was out of date.

It is also worth mentioning this issue that probably should be handled in this PR too: https://github.com/spacetelescope/drizzlepac/issues/1722


perrygreenfield:
Could there be an option to fold the pixmap computation inside the the call to `add_image`? This would be useful in cases where the pixmap is very large and would allow the method to do it for subarrays in cases of very large images (and complex GWCS models).

mcara:
> Could there be an option to fold the pixmap computation inside the the call to `add_image`? This would be useful in cases where the pixmap is very large and would allow the method to do it for subarrays in cases of very large images (and complex GWCS models).

This is one possible approach to deal with large images. There also could be a different function (i.e., `add_block()`) that could deal with blocks of input images or caller could provide a callback, etc. I believe we should add parameters as we introduce the feature that needs those parameters. So, we could possibly add this later when we will support subarrays if this will be deemed to be the best approach.

The beauty of using `pixmap` is that it is completely independent on how WCS are computed and also keep in mind that `pixmap` size is the same as that of the input image and not of the output image and so maybe it is not that critical to be able to split input images into chunks.

drizzle/resample.py
```diff
@@ -0,0 +1,404 @@
...
+            ``"INDEF"`` and ``out_sci`` is **not provided**, the values of
+            ``out_sci`` will be initialized to `numpy.nan`. If ``fillval``
+            is a string that can be converted to a number, then the output
+            pixels with no contributions from input images will be set to this
+            ``fillval`` value.
+
+        out_sci : 2d array of float32, optional
+            A 2D numpy array containing the output image produced by
+            drizzling. On the first call it should be set to zero.
+            Subsequent calls it will hold the intermediate results
```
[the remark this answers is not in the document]
mcara: This is old documentation that applies to the "traditional" workflow in which we start with an image that is all `0`s or `NaN`s and keep adding inputs as needed. Initial  `out_sci` *does not* have to be all `0`s: it can be the previous "end result" upon which we can resample more inputs.

drizzle/resample.py
```diff
@@ -0,0 +1,404 @@
...
+        )
+        self._cversion = _vers  # TODO: probably not needed
+
+        return nmiss, nskip
+
+
+def blot_image(data, pixmap, pix_ratio, exptime, output_pixel_shape,
+               interp='poly5', sinscl=1.0):
+    """
+    Resample input image using interpolation to an output grid.
```
perrygreenfield, line 362: I find this confusing in that the term input seems ambiguous. Is that the input of the original image, or the input of this function. I would start saying that this is intended to do the inverse resampling that drizzle does.

drizzle/resample.py
```diff
@@ -0,0 +1,404 @@
...
+    sampled** back to the pixel coordinate system of input (with distorted
+    WCS) images of ``Drizzle.add_image``.
+
+    Parameters
+    ----------
+
+    data : 2D array
+        Input numpy array of the source image in units of 'cps'.
+
+    pixmap : 3D array
```
perrygreenfield, line 374: Is this the same array as used for drizzle or one that maps the other direction?

drizzle/resample.py
```diff
@@ -0,0 +1,404 @@
...
+        The world coordinate system to resample on.
+
+    output_pixel_shape : tuple of int
```
perrygreenfield, line 377: Similar ambiguity here.

perrygreenfield (commented):
Generally looks very good, but I have some comments.

drizzle/utils.py
```diff
@@ -0,0 +1,87 @@
...
+    pixmap = np.dstack([x, y])
+    if estimate_pixel_scale_ratio:
+         pscale_ratio = (_estimate_pixel_scale(wcs_to, refpix_to) /
+                         _estimate_pixel_scale(wcs_from, refpix_from))
+         return pixmap, pscale_ratio
+
+    return pixmap
+
+
+def _estimate_pixel_scale(wcs, refpix):
```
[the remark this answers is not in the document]
mcara: Yes, as I mentioned, this PR is at "conceptual" stage at this moment. Just to get an idea that proposed API changes look OK. More comments will be added. Documentation was only partially updated. For some parameters it is wrong.

drizzle/utils.py
```diff
@@ -0,0 +1,87 @@
...
+    # estimate pixel scale (in rad) using approximate algorithm
+    # from https://trs.jpl.nasa.gov/handle/2014/40409
+    if refpix is None:
+        if wcs.bounding_box is None:
+            refpix = np.ones(wcs.pixel_n_dim)
```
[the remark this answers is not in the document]
mcara: Any pixel could work. But you are right: more conventional is to choose the center. I will update this

drizzle/resample.py
```diff
@@ -0,0 +1,404 @@
+"""
+The `drizzle` module defines the `Drizzle` class, for combining input
+images into a single output image using the drizzle algorithm.
+"""
+import numpy as np
+
+from . import cdrizzle
+
+
+SUPPORTED_DRIZZLE_KERNELS = [
```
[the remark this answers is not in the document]
mcara: Yes, since no one approved on that JIRA issue.

drizzle/utils.py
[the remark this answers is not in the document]
mcara: That's of course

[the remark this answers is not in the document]
mcara: > more conventional is to choose the center

That's of course if one knows what the bounding box is.

drizzle/resample.py
```diff
@@ -0,0 +1,404 @@
...
+            else:
+                shapes.add(out_ctx.shape)
+
+        if out_shape is None:
+            if not shapes:
+                raise ValueError(
+                    "'out_shape' cannot be None when all output arrays are "
+                    "None."
+                )
+            out_shape = shapes[0]
```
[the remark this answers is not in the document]
mcara: This is a mistake. In all other places I used `pop`. Will be fixed in the next commit.

drizzle/resample.py
mcara: these are comments from the existing code. I did not modify them at this moment. I only moved the `blot_image` out of the `Drizzle` class as it does not belong to drizzle logically and does not need to be in a class.

I do not know who came up with the term `blot` but it is definitely not me.

> this is intended to do the inverse resampling that drizzle does.

I disagree. I think there is nothing "inverse" in here. Resampling is just resampling no matter the direction. Drizzle mixes values from several input pixels into one output pixel. The term "inverse" would suggest un-mingling those values to get the input (to drizzle) image, like Lucy-Richardson deconvolution.

drizzle/resample.py
mcara: it is "inverse" map: it maps from the `wcs_from` to `wcs_to`.

drizzle/resample.py
mcara: similar to what?

drizzle/resample.py
perrygreenfield: Well, I think that an inverse mapping is tightly associated with the term "blot" (I'm not sure why it is called that either). Perhaps a round tripping is the right term, where it is understood that it is not a pure inverse. So I think my comment stands to make the issue clearer since those using blot are almost always using the inverse mapping, aren't they? If you got rid of `blot` and forced people to use drizzling directly, then this wouldn't be an issue. But there is a history to this term.

mcara: Resampling algorithms (interpolating, like "blot", or drizzle) are all "one way" in the sense that you cannot apply it twice with "inverse mapping" (i.e., inverted pixmap) and expect things to "round-trip". Round-tripping will work with "blot" but only if image features can be perfectly (locally) modeled by linear/poly3/poly5 functions. With these kinds of images, in principle, we could use blot to resample from one frame to another and back. Reverse transformation in this case would be performed by _exactly the same_ "blot" algorithm (and code). The only difference is in pixmap. Pixmap is always from the original (image's) coordinate system to the output (destination) coordinate system. 

In any case,
```
I would start saying that this is intended to do the inverse resampling that drizzle does.
```
IMO is fundamentally incorrect: one cannot do "inverse resampling" of drizzled images using interpolation (possibly except for point kernel in the absence of distortions, etc.)

mcara: "inverse resampling" suggests inverse operation to drizzle which is not what blot is doing. "blot" can be used both for forward (meaning resampling a distorted image onto an undistorted frame) transform and backward transform (from the distortion-corrected frame to the distorted frame). We usually use it for the "backward" transformation but it is not the same as "inverse": the final image is not the initial image used as input to drizzle.

perrygreenfield: I think we are talking at cross-purposes. What I mean is that blot is usually taken as mapping in the opposite direction to the original sampling. It is not an inverse of the original resampling, for sure. The point I was making was that there is ambiguity as to what input means, i.e., the original input image used for the original resampling, or the product in the destination pixel coordinate system to be mapped back to the original pixel coordinate system. In other words, with blot one could be confused as to which array is being referred to. It would be nice to clarify that in this case the input is the image generated from the original resampling (or whatever words describe that best). Maybe the current wording is obvious to everyone else, but it wasn't for me. Perhaps we should take a vote :-)

perrygreenfield (approved):
Aside from the comment on one of docstrings, this looks good. I don't want to hold that up due to that comment, which can always be addressed (or not) later.

drizzle/resample.py
```diff
@@ -0,0 +1,406 @@
...
+        plane_no, id_in_plane = self._increment_ctx_id()
+
+        # Ensure that the fillval parameter gets properly interpreted
+        # for use with tdriz
+        if in_units == 'cps':
+            expscale = 1.0
+        else:
+            expscale = exptime
+
+        self.exptime += exptime
```
nden, line 278: Is this used anywhere?

drizzle/utils.py
```diff
@@ -0,0 +1,99 @@
...
+    pixel_scale_ratio : float
+        Estimate the ratio of "to" to "from" WCS pixel scales. This value is
+        returned only when ``estimate_pixel_scale_ratio`` is `True`.
+
+    """
+    # We add one to the pixel co-ordinates before the transformation and subtract
+    # it afterwards because wcs co-ordinates are one based, while pixel co-ordinates
+    # are zero based, The result is the final values in pixmap give a tranformation
+    # between the pixel co-ordinates in the first image to pixel co-ordinates in the
+    # co-ordinate system of the second.
```
nden, line 64: Should the above comment be deleted?

drizzle/utils.py
```diff
@@ -0,0 +1,99 @@
...
+
+    if wcs_from.pixel_shape is None:
+        raise ValueError('The "from" WCS must have pixel_shape property set.')
+    y, x = np.indices(wcs_from.pixel_shape, dtype=np.float64)
+    x, y = wcs_to.invert(*wcs_from(x, y))
```
nden, line 69: I wonder if it's better to use the common API here instead of using gwcs.WCS methods.

drizzle/resample.py
mcara: No. It is an output for the user.

drizzle/utils.py
mcara: Of course it will be deleted. At this stage this PR is about API only. However, since it looks like you and @perrygreenfield approve this API redesign, I will now work on cleaning up comments and docstrings.

drizzle/resample.py
```diff
@@ -0,0 +1,406 @@
...
+        if out_wht is not None:
+            out_wht = np.asarray(out_wht, dtype=np.float32)
+            shapes.add(out_wht.shape)
+
+        if out_ctx is not None:
+            out_ctx = np.asarray(out_ctx, dtype=np.int32)
+            if len(out_ctx.shape) == 3:
+                shapes.add(out_ctx.shape[1:])
+            else:
+                shapes.add(out_ctx.shape)
```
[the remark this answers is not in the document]
mcara: 🤦‍♂️

mcara:
Section **"The Drizzle Library"** in the `README.rst` file has been removed because code examples are no longer valid/up-to-date with API changes. Updated examples should be provided either as a separate commit in this PR or a separate PR.

perrygreenfield (approved):
Just a minor question that shouldn't affect approval.

drizzle/tests/test_drizzle.py
```diff
@@ -110,7 +110,7 @@ def centroid_statistics(title, fname, image1, image2, amp, size):
     distances = centroid_distances(image1, image2, amp, size)
     indexes = (0, int(len(distances) / 2), len(distances) - 1)
     fd = open(fname, 'w')
-    fd.write("*** %s ***\n" % title)
```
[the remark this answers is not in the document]
mcara: it is not necessary but doing so is closest to the original code and probably is not needed, especially for the `s` format (`f` is different though).

drizzle/utils.py
mcara: I forgot about this. will be fixed

mcara:
I wonder whether https://github.com/spacetelescope/drizzlepac/pull/1767 should be ported to drizzle too (it is basically to get beautiful images with a fraction of pixels having potentially unusable values for science). Basically, instead of having one constant `fillval`, let's fill pixels with whatever we can get from input images even if those input pixels have 0 weight. It could be an option...

src/cdrizzleapi.c
```diff
@@ -575,7 +568,7 @@ clip_polygon_wrap(PyObject *self, PyObject *args)
 
 static struct PyMethodDef cdrizzle_methods[] = {
     {"tdriz",  (PyCFunction)tdriz, METH_VARARGS|METH_KEYWORDS,
-    "tdriz(image, weight, output, outweight, context, uniqid,  xmin, ymin, scale, pfract, kernel, inun, expin, wtscl, fill, nmiss, nskip, pixmap)"},
+    "tdriz(image, weights, pixmap, output, counts, context, uniqid,  xmin, xmax, ymin, ymax, scale, pixfrac, kernel, in_units, expscale, wtscale, fillstr)"},
```
nden, line 571: I'm not sure I understand this change. It looks like you are using here the kw list, not the input to the PyArg_ParseTupleAndKeywords function

mcara: This is just the docs string that documents the call signature for this function from Python and the actual order in which input arguments to `tdriz` will be parsed by the `PyArg_ParseTupleAndKeywords`:
```python
In [1]: from drizzle import cdrizzle

In [2]: ?cdrizzle.tdriz
Docstring: tdriz(image, weights, pixmap, output, counts, context, uniqid,  xmin, xmax, ymin, ymax, scale, pixfrac, kernel, in_units, expscale, wtscale, fillstr)
Type:      builtin_function_or_method
```

`tblot` needs a similar docstring update.

mcara: For example, old docstring was showing just these args: `xmin, ymin` and the correct/actual list is `xmin, xmax, ymin, ymax`

nden (approved):

mcara:
Regression tests using this branch and `jwst` pipeline `main` are running here: https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/1801/

mcara:
Re-running regression tests after most recent commits:

JWST: https://github.com/spacetelescope/RegressionTests/actions/runs/11473268798
Roman: https://github.com/spacetelescope/RegressionTests/actions/runs/11473286488

mcara:
Regression tests agains released version of Roman pipeline:

With released `drizzle`: https://github.com/spacetelescope/RegressionTests/actions/runs/11483667508
With this PR: https://github.com/spacetelescope/RegressionTests/actions/runs/11482465914

Both tests seem to have the same 28 failures, mostly in dar current tests and related to missing data files (probably regression tests have evolved meanwhile).

PR #150, opened 2024-07-25
Fix return type of PyInit_cdrizzle

olebole:
The return type of the init function is `PyMODINIT_FUNC`, not `PyMODINIT_FUNC *`.

This was found by an error raised when built with gcc-14, see [Debian#1075414](https://bugs.debian.org/1075414).


mcara (approved):
Thank you! LGTM

PR #151, opened 2024-08-09
allow `context` to be `None`

braingram:
This PR allows `context` to be `None` for `tdriz` (to prevent the need for a context array).

One place where this would be useful is within the jwst and romancal pipelines where drizzle is used to generate intermediate products for outlier detection. These operations can involve large arrays and using `context=None` (and not generating a large context array) can result in memory savings.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,12 @@
 Release Notes
 =============
 
-1.15.2 (2025-06-17)
+1.15.3 (unreleased)
+===================
+
+- allow ``context`` to be ``None`` in ``tdriz`` [#151]
+
+1.15.2 (2024-06-17)
 ===================
 
 - build wheels with Numpy 2.0 release candidate [#149]
--- a/drizzle/tests/test_drizzle.py
+++ b/drizzle/tests/test_drizzle.py
@@ -796,3 +796,39 @@ def test_flux_conservation_distorted(kernel):
         atol=0.0,
         rtol=0.0001,
     )
+
+
+def test_no_context_array():
+    """
+    Test that providing None for "outctx" is supported.
+    """
+    # initialize input:
+    insci = np.ones((200, 400), dtype=np.float32)
+    inwht = np.ones((200, 400), dtype=np.float32)
+    inwht[:, 150:155] = 0
+
+    # initialize output:
+    outsci = np.zeros((210, 410), dtype=np.float32)
+    outwht = np.zeros((210, 410), dtype=np.float32)
+    outctx = None
+
+    # define coordinate mapping:
+    pixmap = np.moveaxis(np.mgrid[1:201, 1:401][::-1], 0, -1)
+
+    # resample:
+    cdrizzle.tdriz(
+        insci, inwht, pixmap,
+        outsci, outwht, outctx,
+        uniqid=1,
+        xmin=0, xmax=400,
+        ymin=0, ymax=200,
+        pixfrac=1,
+        kernel='square',
+        in_units='cps',
+        expscale=1,
+        wtscale=1,
+        fillstr='INDEF',
+    )
+
+    # check that no pixel with 0 weight has any counts:
+    assert np.sum(np.abs(outsci[(outwht == 0)])) == 0.0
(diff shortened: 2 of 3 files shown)


braingram:
2 failures in CI:
- build: only failing for python 3.13 on windows during what looks like an attempt to build numpy (likely due to a lack of a released 3.13 wheel)
- patch coverage: failing due to the `dodrizzle` change. If preferred I'm happy to drop those changes from this PR. It's not required by jwst or romancal and it appears that the `dodrizzle` function is untested.

mcara:
I do not think this PR is needed for the current API. For the new proposed API, I have already made a commit to add this option although turning off of CTX is done entirely in Python: https://github.com/spacetelescope/drizzle/pull/134/commits/e0132a210ca2d8c2e31d8e791aa6ee52701882c4

mcara:
Also, as we have discussed in a meeting, ultimately C code should be modified to not even need a temporary context array. This will be done at a later date in a different PR.

braingram:
> Also, as we have discussed in a meeting, ultimately C code should be modified to not even need a temporary context array. This will be done at a later date in a different PR.

Thanks for taking a look.  Is there an issue with the modifications in this PR? It already modifies the c code to not require a temporary array.

mcara:
Oh, I didn't realize the C code would work with a context array set to NULL. I just looked at the code and it should work with proposed changes. Any objection to change `skip_context` to `no_ctx`?

braingram:
Good idea! I'll update the PR with the new name on Monday.

mcara:
@braingram I was actually wondering whether `disable_ctx`, `disable_var(err)`, etc. or `enable_ctx/var/err...` wouldn't be better than noth `skip_` and `no_`. I think it is no clear to use either `enable` or `disable`. If so, maybe we should use `enable` and set it to `False` by default and explicitly turn on processing of CTX, VAR, ...

braingram:
I was hoping to keep this backwards compatible and minimal (to get it out sooner rather than later). Setting the default context processing to `False` would be an issue for jwst and romancal (since they both assume it's `True`).

One other option which might work would be to make the context argument optional, set it's default in c to `NULL`, then only convert it to a ndarray if it's not `NULL`. I think that could work for jwst and romancal (which could just not provide a `context` argument to disable context array handling).

@mcara would you like me to update this PR to try defaulting `context` to `NULL` (so we don't need to add another argument) or to rename the argument to something else (perhaps `disable_context` defaulting to `False`)?

mcara:
> I was hoping to keep this backwards compatible and minimal (to get it out sooner rather than later). Setting the default context processing to `False` would be an issue for jwst and romancal (since they both assume it's `True`).

Yes, leave the default to True. I thought that this will go with some redesign of the resample step and if so change in default would not matter (and my thinking was that it would be better to _explicitly_ turn on additional processing when needed). On the other hand, if this is urgent, I assume pipelines will need to get updated anyway. Still, it is safer to leave this on by default.

mcara:
@braingram I think I understood what you meant at the meeting. Yes, I think we do not need this argument at all for the `tdriz` and `dodrizzle` functions. We should turn off context array if input argument for the context array is `None`.

On the other hand, this `enable_context` argument is needed for the `Drizzle` class, specifically to modify this code:
https://github.com/spacetelescope/drizzle/blob/dfdd16bf735e924f495acd134c68c40f07fcb6c2/drizzle/drizzle.py#L175-L176
to leave `outcon` to be `None` when `enable_context=False`.

braingram:
Thanks for giving this another look. I removed the `dodrizzle` changes so now the PR only modifies `tdriz`.

Does it work for you to keep the `dodrizzle` and `Drizzle` class changes to a separate PR or perhaps not updating them? Skimming https://github.com/spacetelescope/drizzle/pull/134/ these appear to be removed in the redesign.

mcara (commented):
Can you add a changelog entry, please?

braingram:
I added a changelog in [c8e0ed8](https://github.com/spacetelescope/drizzle/pull/151/commits/c8e0ed862b882050010a67e79ad61ce1f985d213)

I also fixed the year for the last release changing it from 2025 to 2024. I didn't know you could time travel :)

mcara:
> I also fixed the year for the last release changing it from 2025 to 2024. I didn't know you could time travel :)

Some of us (@zacharyburnett) do, https://github.com/spacetelescope/drizzle/commit/9ecb4b7889b6fb9951cef87f96c83aa2e7a27d00

mcara (approved):
LGTM

PR #156, opened 2024-08-19
allow python 3.13

braingram:
This PR:
- adds python 3.13 testing to the CI: https://github.com/spacetelescope/drizzle/actions/runs/10473588949/job/29005811343?pr=156
- allows python 3.13 in pyproject.toml
- re-enables to previously disabled tox devdeps test https://github.com/spacetelescope/drizzle/actions/runs/10473588949/job/29005812771?pr=156

I added the devdeps change only because I saw the comment:
https://github.com/spacetelescope/drizzle/blob/f8ea30dc5b1eda828aa1e4191c9d9205f01551ae/.github/workflows/ci.yml#L35-L36
Since numpy 2.0 is out I figured I'd see if it works again (which it did). Let me know if it's preferable to have this change in a different PR.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,7 +1,7 @@
 [tox]
 env_list =
     check-{style,security}
-    py{39,310,311,312,313}{,-xdist}{,-devdeps,-cov}
+    py{310,311,312,313}{,-xdist}{,-devdeps,-cov}
 
 # tox environments are constructed with so-called 'factors' (or terms)
 # separated by hyphens, e.g. test-devdeps-cov. Lines below starting with factor:
--- a/tox.ini
+++ b/tox.ini
@@ -1,7 +1,7 @@
 [tox]
 env_list =
     check-{style,security}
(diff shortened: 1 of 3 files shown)


braingram:
Now python 3.13 is passing... I'm inclined to say we leave the `<3.13` pin in place for the next release and remove it only when it's enabled in the tests.

tox.ini
```diff
@@ -1,7 +1,7 @@
 [tox]
 env_list =
     check-{style,security}
-    py{39,310,311,312}{,-xdist}{,-devdeps,-cov}
+    py{39,310,311,312,313}{,-xdist}{,-devdeps,-cov}
```
mcara, line 4: Shouldn't 39 be removed from the list given  the constrain from `pyproject.toml`?

braingram, line 4:
```suggestion
    py{310,311,312,313}{,-xdist}{,-devdeps,-cov}
```
Good call.

mcara (approved):

PR #161, opened 2024-10-25
Use astropy pixel_to_pixel in calc_pixmap

### LINKED ISSUE #51: Use astropy pixel_to_pixel instead of internal calc_pixmap function
Use `astropy.wcs.utils.pixel_to_pixel` instead of

https://github.com/spacetelescope/drizzle/blob/291b8b6249f2b24cac801ea99b383f7bf61dd1aa/drizzle/calc_pixmap.py#L5

if possible.

This will require bumping up the `astropy` and other dependency versions.

mcara:
Closes #51


jdavies-st (approved):
Looks good.  I guess I had the thought when I opened the issue that it would completely replace `calc_pixmap`, but I see that one still needs to bundle it into a pixmap that the drizzle C code can use.  Bummer.

Anyway, looks good.  Hopefully `jwst.resample` can now use the `calc_pixmap` function here.  I recall it had been copied over to `jwst.resample` because for along time the one here did not support `gwcs` and APE 14 did not exist.

👍

nden:
This will need careful consideration, and as a minimum a regression test run. Looking at the astropy `pixel_to_pixel` function, it uses the high level API `pixel_to_world` which generates `SkyCoord` objects. It's supposed to be efficient with views of arrays. All I'm saying is it needs some more testing specifically for the use in drizzle.

mcara:
There are unit tests testing `calc_pixmap` in `drizzle`.  `calc_pixmap` is not used in the pipelines and so running regression tests would not test this helper function.

braingram:
> There are unit tests testing `calc_pixmap` in `drizzle`. `calc_pixmap` is not used in the pipelines and so running regression tests would not test this helper function.

How would you like to coordinate this with https://github.com/spacetelescope/stcal/pull/320 (which will start using `calc_pixmap` in stcal and the pipelines)?

mcara:
This cannot be merged because after https://github.com/spacetelescope/jwst/pull/9404 NIRSPEC WCS no longer works with `separability_matrix` - see  https://github.com/astropy/astropy/issues/18523.

So we either wait for `separability_matrix()` to be fixed or modify NIRSPEC WCS to propagate name in a different way that doesn't crash `separability_matrix()` used by `pixel_to_pixel()`.

PR #162, opened 2024-10-25
add codecov token to uploader step

zacharyburnett:
this should fix the codecov upload not working for `main` (it works without a token on PRs)


mcara (approved):
Thanks!

PR #163, opened 2024-12-07
Add support for co-adding with squared weights

mcara:
This PR adds support for resampling and adding resampled (error, variance) images with squared weights. This is to facilitate standard error propagation.

### DIFF
--- a/drizzle/resample.py
+++ b/drizzle/resample.py
@@ -450,7 +450,12 @@ class Drizzle:
             self._out_img = out_img
 
     def _alloc_output_arrays2_init(self, out_img2=None):
-        assert self._out_img2 is None
+        if hasattr(self, "_out_img2") and self._out_img2 is not None:
+            raise AssertionError(
+                "It is expected that _alloc_output_arrays2_init is called "
+                "before Drizzle._out_img2 is set."
+            )
+
         if out_img2 is None:
             return
 
--- a/drizzle/resample.py
+++ b/drizzle/resample.py
@@ -37,19 +37,19 @@ class Drizzle:
     pixel. This is accomplished via *context image*.
 
     Main outputs of :py:meth:`add_image` can be accessed as class properties
-    ``out_img``, ``out_wht``, ``out_ctx``, and ``exptime``.
+    ``out_img``, ``out_img2``, ``out_wht``, ``out_ctx``, and ``exptime``.
 
     .. warning::
-        Output arrays (``out_img``, ``out_wht``, and ``out_ctx``) can be
-        pre-allocated by the caller and be passed to the initializer or the
-        class initializer can allocate these arrays based on other input
+        Output arrays (``out_img``, ``out_img2``, ``out_wht``, and ``out_ctx``)
+        can be pre-allocated by the caller and be passed to the initializer or
+        the class initializer can allocate these arrays based on other input
         parameters such as ``output_shape``. If caller-supplied output arrays
-        have the correct type (`numpy.float32` for ``out_img`` and ``out_wht``
-        and `numpy.int32` for the ``out_ctx`` array) and if ``out_ctx`` is
-        large enough not to need to be resized, these arrays will be used as is
-        and may be modified by the :py:meth:`add_image` method. If not,
-        a copy of these arrays will be made when converting to the expected
-        type (or expanding the context array).
+        have the correct type (`numpy.float32` for ``out_img``, ``out_img2``
+        and ``out_wht`` and `numpy.int32` for the ``out_ctx`` array) and if
+        ``out_ctx`` is large enough not to need to be resized, these arrays
+        will be used as is and may be modified by the :py:meth:`add_image`
+        method. If not, a copy of these arrays will be made when converting
+        to the expected type (or expanding the context array).
 
     Output Science Image
     --------------------
@@ -165,9 +165,10 @@ class Drizzle:
 
     """
 
-    def __init__(self, kernel="square", fillval=None, out_shape=None,
-                 out_img=None, out_wht=None, out_ctx=None, exptime=0.0,
-                 begin_ctx_id=0, max_ctx_id=None, disable_ctx=False):
+    def __init__(self, kernel="square", fillval=None, fillval2=None,
+                 out_shape=None, out_img=None, out_wht=None, out_ctx=None,
+                 out_img2=None, exptime=0.0, begin_ctx_id=0, max_ctx_id=None,
+                 disable_ctx=False):
         """
         kernel: str, optional
             The name of the kernel used to combine the input. The choice of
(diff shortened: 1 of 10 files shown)


src/cdrizzleapi.c
```diff
@@ -610,9 +926,10 @@ clip_polygon_wrap(PyObject *self, PyObject *args) {
 
 static struct PyMethodDef cdrizzle_methods[] = {
     {"tdriz", (PyCFunction)tdriz, METH_VARARGS | METH_KEYWORDS,
-     "tdriz(image, weights, pixmap, output, counts, context, uniqid, xmin, "
-     "xmax, ymin, ymax, scale, pixfrac, kernel, in_units, expscale, wtscale, "
-     "fillstr)"},
+     "tdriz(image, weights, pixmap, output, counts, context, image2, "
+     "counts2, "
```
melanieclarke, line 930: I don't see counts2 anywhere else - is it supposed to be here?

src/cdrizzlebox.c
```diff
@@ -307,13 +344,20 @@ do_kernel_point(struct driz_param_t *p) {
...
 
                 } else {
-                    vc = get_pixel(p->output_counts, ii, jj);
-
                     /* Allow for stretching because of scale change */
                     d = get_pixel(p->data, i, j) * scale2;
+                    if (d2) {
+                        for (k = 0; k < ndata2; ++k) {
+                            if (p->data2[k]) {
+                                d2[k] = get_pixel(p->data2[k], i, j) * scale2;
```
melanieclarke, line 352: I think this should be scale2 squared

src/cdrizzlebox.c
```diff
@@ -419,6 +491,15 @@ do_kernel_gaussian(struct driz_param_t *p) {
 
                 /* Allow for stretching because of scale change */
                 d = get_pixel(p->data, i, j) * scale2;
+                if (d2) {
+                    for (k = 0; k < ndata2; ++k) {
+                        if (p->data2[k]) {
+                            d2[k] = get_pixel(p->data2[k], i, j) * scale2;
```
melanieclarke, line 497: scale2 squared here too

src/cdrizzlebox.c
```diff
@@ -553,6 +659,15 @@ do_kernel_lanczos(struct driz_param_t *p) {
...
```
melanieclarke, line 665: scale2 squared here too

src/cdrizzlebox.c
```diff
@@ -687,7 +827,16 @@ do_kernel_turbo(struct driz_param_t *p) {
                 nhit = 0;
 
                 /* Allow for stretching because of scale change */
-                d = get_pixel(p->data, i, j) * (float)scale2;
+                d = get_pixel(p->data, i, j) * scale2;
+                if (d2) {
+                    for (k = 0; k < ndata2; ++k) {
+                        if (p->data2[k]) {
+                            d2[k] = get_pixel(p->data2[k], i, j) * scale2;
```
melanieclarke, line 834: scale2 squared here too

src/cdrizzlebox.c
```diff
@@ -830,6 +1004,15 @@ do_kernel_square(struct driz_param_t *p) {
 
             /* Allow for stretching because of scale change */
             d = get_pixel(p->data, i, j) * scale2;
+            if (d2) {
+                for (k = 0; k < ndata2; ++k) {
+                    if (p->data2[k]) {
+                        d2[k] = get_pixel(p->data2[k], i, j) * scale2;
```
melanieclarke, line 1010: scale2 squared here too

melanieclarke (commented):
A couple questions and suggestions below, but in particular: the weights and area overlaps look like they are squared in the right places, but I think the pixel scale also needs to be squared before applying it to the squared data arrays.

src/cdrizzlebox.c
mcara: Yeah, I thought of this but then forgot to fix it (due to other things I needed to fix). Thanks for reminding

src/cdrizzleapi.c
mcara: no, no need for `counts2`. Just square `counts`

drizzle/resample.py
```diff
@@ -554,6 +642,32 @@ def add_image(self, data, exptime, pixmap, scale=1.0,
...
+                for k, d in enumerate(data2):
+                    if d is None or d.size == 0:
+                        data2[k] = None
+                    else:
+                        shapes2.add(d.shape)
+
+                if len(shapes2) > 1 and shapes2.pop() != data.shape:
+                    raise ValueError(
+                        "'data2' shape(s) is not consistent with 'data' shape."
+                    )
```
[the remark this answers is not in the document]
mcara: > Why set data2 to None here? 

Do you mean `data2[k] = None` on line 660? If so, that's being done when the error/variance array are empty. Setting to `None` allows code to not resample that variance array.

> And it looks like you're only checking one of the data shapes in shapes2 against data.shape - is that what was intended?

`shapes2` is a set and all input arrays should have all the same shape. Appending the same `ndarray.shape` to a set will result in a set with only one element. Yes, it is needed because JWST pipeline has unit tests that try to resample variance arrays of different shapes from the `data` array.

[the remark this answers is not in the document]
melanieclarke: Oh, okay, to handle when d.size == 0?  That makes sense.

For shapes2, what I meant to ask was - what if data2 is a list with arrays of different shapes, so the set has more than one element, but the first one popped matches the input data?  Then it won't raise the ValueError.  Maybe you meant 'or' instead of 'and'?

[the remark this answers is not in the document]
mcara: > Maybe you meant 'or' instead of 'and'?

Yes, this line needs to be fixed

src/cdrizzleapi.c
melanieclarke: I think this one reference to counts2 still needs to be removed.  Other than that, all your changes look good to me.

mcara: Done

melanieclarke (approved):
Everything looks good now, from my perspective!

src/cdrizzleapi.c
```diff
@@ -48,16 +48,160 @@ scale_image(PyArrayObject *image, double scale_factor) {
  * Top level function for drizzling, interfaces with python code
  */
 
+static int
+process_array_list(PyObject *list, integer_t *nx, integer_t *ny,
```
kmacdonald-stsci, line 52: Add comment to document what this function does, as well as what the function parameters are.

src/cdrizzleapi.c
```diff
@@ -48,16 +48,160 @@ scale_image(PyArrayObject *image, double scale_factor) {
...
+
+_exit_on_err:
+
+    Py_XDECREF(list_elem);
+    if (arr_list) {
+        for (i = 0; i < *nmax; ++i) {
+            Py_XDECREF(arr_list[i]);
+        }
+        free(arr_list);
+    }
```
kmacdonald-stsci, line 188: Needs a return value.

src/cdrizzleapi.c
```diff
@@ -48,16 +48,160 @@ scale_image(PyArrayObject *image, double scale_factor) {
...
+}
+
 static PyObject *
 tdriz(PyObject *obj UNUSED_PARAM, PyObject *args, PyObject *keywords) {
```
kmacdonald-stsci, line 192: This function roughly 400 lines long and should be broken down into smaller sub-functions.

src/cdrizzleapi.c
```diff
@@ -228,47 +536,38 @@ tdriz(PyObject *obj UNUSED_PARAM, PyObject *args, PyObject *keywords) {
     p.exposure_time = expin;
     p.weight_scale = wtscl;
     p.fill_value = fill_value;
+    p.fill_value2 = fill_value2;
```
kmacdonald-stsci, line 539: One letter variable names should be avoided for things other than indices.  I recommend changing `p` to a more descriptive variable name.

kmacdonald-stsci (approved):
Minor comments, but otherwise looks good.

mcara:
Regression tests:

JWST: https://github.com/spacetelescope/RegressionTests/actions/runs/17073575486
Roman: https://github.com/spacetelescope/RegressionTests/actions/runs/17073538185

PR #167, opened 2025-01-07
Add unit tests that use gwcs. Fix how the bounding box is handled.

mcara:
Adds unit tests that utilize `gwcs` in addition to FITS WCS.

### DIFF
--- a/drizzle/utils.py
+++ b/drizzle/utils.py
@@ -61,8 +61,17 @@ def calc_pixmap(wcs_from, wcs_to, shape=None):
         shape = wcs_from.array_shape
         if shape is None:
             if (bbox := getattr(wcs_from, "bounding_box", None)) is not None:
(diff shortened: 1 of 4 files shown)


drizzle/utils.py
```diff
@@ -61,6 +61,8 @@ def calc_pixmap(wcs_from, wcs_to, shape=None):
         shape = wcs_from.array_shape
         if shape is None:
             if (bbox := getattr(wcs_from, "bounding_box", None)) is not None:
+                if not isinstance(bbox, tuple):
+                    bbox = bbox.bounding_box()
```
nden, line 65: This will work only if bounding_box is an instance of `modeling.bounding_box.ModelBoundingBox`, so a better check is that it is that instance.

nden (approved):
Looks good, other than the one comment

PR #175, opened 2025-02-28
Revert computation of pixel overlap for square kernel, cache some computations

t-brandt:
This PR covers https://jira.stsci.edu/browse/JP-3902.  By reverting to the previous function called "boxer" for calculating the overlap between input and output pixels with the square kernel, we can cut down the runtime to of the resample step to 2/3 of its current value.  We just need to compute and cache the slope, inverse slope, and direction of each line segment.  Logic within the resurrected function "sgarea" is simplified significantly, and the number of floating point divisions is reduced to eight per quadrilateral (two per vertex).  In my tests, results on the JWST Pipeline Notebooks differ at the level expected from floating point roundoff differences, with nanstd a few 1e-10 times nanmedian.  I have left compute_area for now (it is covered by tests), and we can probably adapt old tests for boxer and sgarea.

### DIFF
--- a/src/cdrizzlebox.c
+++ b/src/cdrizzlebox.c
@@ -204,144 +204,8 @@ boxer(double is, double js,
 }
 
 /** ---------------------------------------------------------------------------
- * Compute area of box overlap. Calculate the area common to input clockwise
- * polygon x(n), y(n) with square (is, js) to (is+1, js+1). This version is for
- * a quadrilateral. Used by do_square_kernel.
- *
- * is: x coordinate of a pixel on the output image
- * js: y coordinate of a pixel on the output image
- * x:  x coordinates of endpoints of quadrilateral containing flux of input
- * pixel y:  y coordinates of endpoints of quadrilateral containing flux of
- * input pixel
- */
-
-double
-compute_area(double is, double js, const double x[4], const double y[4]) {
-    int ipoint, jpoint, idim, jdim, iside, outside, count;
-    int positive[2];
-    double area, width;
-    double midpoint[2], delta[2];
-    double border[2][2], segment[2][2];
-
-    /* The area for a qadrilateral clipped to a square of unit length whose
-     * sides are aligned with the axes. The area is computed by computing the
-     * area under each line segment clipped to the boundary of three sides of
-     * the sqaure. Since the computed width is positive for two of the sides and
-     * negative for the other two, we subtract the area outside the
-     * quadrilateral without any extra code.
-     */
-    area = 0.0;
-
-    border[0][0] = is - 0.5;
-    border[0][1] = js - 0.5;
-    border[1][0] = is + 0.5;
-    border[1][1] = js + 0.5;
-
-    for (ipoint = 0; ipoint < 4; ++ipoint) {
-        jpoint = (ipoint + 1) & 03; /* Next point in cyclical order */
-
-        segment[0][0] = x[ipoint];
-        segment[0][1] = y[ipoint];
-        segment[1][0] = x[jpoint];
-        segment[1][1] = y[jpoint];
-
-        /* Compute the endpoints of the line segment that
-         * lie inside the border (possibly the whole segment)
-         */
-
-        for (idim = 0, count = 3; idim < 2; ++idim) {
-            for (iside = 0; iside < 2; ++iside, --count) {
-                delta[0] = segment[0][idim] - border[iside][idim];
-                delta[1] = segment[1][idim] - border[iside][idim];
-
-                positive[0] = delta[0] > 0.0;
-                positive[1] = delta[1] > 0.0;
-
-                /* If both deltas have the same signe there is no baundary
-                 * crossing
-                 */
-                if (positive[0] == positive[1]) {
-                    /* A diagram will convince that you decide a point is
-                     * inside or outside the boundary by the following test
-                     */
-                    if (positive[0] == iside) {
-                        /* Segment is entirely outside the boundary */
-                        if (count == 0) {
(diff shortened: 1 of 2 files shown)


mcara:
Could you, please, rebase this PR

src/cdrizzlebox.c
```diff
@@ -839,22 +968,44 @@ do_kernel_square(struct driz_param_t *p) {
...
+	       division calls later.
+	     */
+
+	    for (ii = 0; ii < 4; ii++) {
+	        dx = xout[(ii+1) & 0x3] - xout[ii];
+	        dy = yout[(ii+1) & 0x3] - yout[ii];
+	        if (dx >= 0)
+		    sgn_dx[ii] = 1;
+	        else
+		    sgn_dx[ii] = -1;
```
mcara, line 983:
```suggestion
	        if (dx >= 0) {
		      sgn_dx[ii] = 1;
		    } else {
		      sgn_dx[ii] = -1;
		    }
```

src/cdrizzlebox.c
```diff
@@ -839,22 +968,44 @@ do_kernel_square(struct driz_param_t *p) {
...
+	        slope[ii] = dy / dx;
+	        inv_slope[ii] = dx / dy;
+	    }
+	    inv_jaco = 1.0 / jaco;
```
mcara, line 987: Maybe just do this:
```suggestion
        inv_jaco = 2.0f / ((xout[1] - xout[3]) * (yout[0] - yout[2]) -
                           (xout[0] - xout[2]) * (yout[1] - yout[3]));
```
and remove `jaco` variable altogether?

mcara (approved):
Overall it looks good. It increases performance of resampling of a single "NIRCAM" image ("add_image" method) in the `test_flux_conservation_distorted_distributed_sources()` test by a factor of 3 (on my machine). Minor comments:

- I know that existing code is often written without using `{}` for single line `if` or `for`, ... But for the new code we should avoid doing this.
- One unused variable.

mcara (approved):
Actually, could you, please, rebase again.

mcara:
Regression tests:
JWST: https://github.com/spacetelescope/RegressionTests/actions/runs/14085678491
Roman: https://github.com/spacetelescope/RegressionTests/actions/runs/14085713398

mcara (changes requested):
Also, let's wait until regression tests pass.

mcara:
Well, there are 11 failing tests in the JWST regression tests all related to MIRI LRS. For reference, here is a regression test for JWST using `main` branch for `jwst` package and released version of `drizzle`:

https://github.com/spacetelescope/RegressionTests/actions/runs/14093632129

All tests pass. Just to be safe, I have also started a new test using `main` branches of `jwst` and `drizzle` packages (to exclude the case if differences are due to something already on `main` (for `drizzle`):

https://github.com/spacetelescope/RegressionTests/actions/runs/14095000135

src/cdrizzlebox.c
```diff
@@ -803,58 +928,80 @@ do_kernel_square(struct driz_param_t *p) {
...
-            }
-
-            /* Work out the area of the quadrilateral on the output grid.
-               Note that this expression expects the points to be in clockwise
-               order */
+	    /* Assuming we don't need to extrapolate, call a more
+	     * efficient interpolator that takes advantage of the fact
+	     * that pixfrac<1 and that we are using a square grid.
+	     */
+	    if (i > 0 && i < mapsize[0] - 1 && j > 0 && j < mapsize[1] - 1) {
```
mcara, line 935:
```suggestion
	    if (i > 0 && i < mapsize[0] - 2 && j > 0 && j < mapsize[1] - 2) {
```

JWST regression tests are failing. My investigation shows that this is the most likely culprit.

mcara (changes requested):

src/cdrizzlebox.c
t-brandt: Good catch, thanks.

mcara:
@t-brandt Sorry for delay, could you please take a look at the regression tests? Unfortunately my suggestion did not help (at least not completely). There is still something that introduces differences in tests

t-brandt:
I looked at some of the regression tests.  Differences between the files are small but larger than the allowed tolerances.  I think they are consistent with expectations from refactoring a calculation in a way that should be equivalent but isn't due to truncation error.  The largest changes tend to be next to NaN pixels, where small differences in pixel weights would be magnified.  The differences here flow down to other data products, causing knock-on failures in the reg tests.  The root of the failures is shown here:

[absreldiff.pdf](https://github.com/user-attachments/files/20096064/absreldiff.pdf)
[value.pdf](https://github.com/user-attachments/files/20096065/value.pdf)

As an aside, I added a -O3 compiler flag.  This significantly improved performance on a linux machine.  I would be happy to use -Ofast if you feel that very small and potentially unpredictable discrepancies are ok, since I don't think these would impact science given the approximate nature of drizzle itself.

mcara:
@t-brandt Ignore my previous message: `O3` was added to non-win32 branch of if (I should have expanded the diff).

mcara:
A new set of regression tests was started here:

- JWST: https://github.com/spacetelescope/RegressionTests/actions/runs/15672468375
- Romancal: https://github.com/spacetelescope/RegressionTests/actions/runs/15672471929

mcara:
@t-brandt One last ask. Can you please re-format C code using https://github.com/spacetelescope/drizzle/blob/main/.clang-format? In VSCode this can be adjusted in settings. I do not know about other editors/IDE.

Thanks!

Issue #189, opened 2025-07-02
All `nan` output from distorted pixmap


braingram:
While testing 2.1.0 I stumbled on a condition where a (possibly) reasonable pixmap produces an all `nan` output.

Testing with 2.0.1 this occurs for both versions.

```python
import numpy as np
from drizzle.resample import Drizzle

pixmap = np.array([
    [
        [0.31, 1.],
        [1.01, 1.],
        [2.01, 1.],
    ],
    [
        [0.31, 0.],
        [1.01, 0.],
        [1.71, 0.],
    ]
], dtype="f8")

in_shape = pixmap.shape[:2]
out_shape = (4, 4)


d0 = Drizzle(
    kernel='square',
    fillval='nan',
    out_shape=out_shape,
    disable_ctx=True,
)

img = np.full(in_shape, 42, dtype=np.float32)

d0.add_image(
    img,
    exptime=11.776,
    in_units='cps',
    pixfrac=1.0,
    pixmap=pixmap,
    scale=1.0,
    wht_scale=1.0,
)

img = d0.out_img
max_value = np.nanmax(img)
min_value = np.nanmin(img)

print(f"Max: {max_value}")
print(f"Min: {min_value}")
print(img)
```
Produces:
```
Max: nan
Min: nan
[[nan nan nan nan]
 [nan nan nan nan]
 [nan nan nan nan]
 [nan nan nan nan]]
```
I would expect something closer to:
```
Max: 42.000003814697266
Min: 41.999996185302734
[[42.       42.       42.             nan]
 [42.       42.       42.             42.      ]
 [      nan       nan       nan       nan]
 [      nan       nan       nan       nan]]
```

mcara:
I take back my previous comment: the shape (form) is correct. I think it is an unusual (singular?) pixel map input pixels with different ordinates map to the same output ordinate. Is there a real transformation behind these values?

mcara:
In any case, I get the same output from both `2.0.1` and `2.1.0` using your pixmap and what I thought is the correct pixmap (with `dstack` - but for a pixmap of shape (3, 2, 2) this does not matter).

braingram:
> > Testing with 2.0.1 this occurs for both versions.
> 
> versions of what?

The all `nan` output is seen for 2.0.1 (from pypi) and 2.1.0 (when it was on pypi). I still see this locally with 2.0.1 from pypi with or without the `dstack`. However if I build drizzle locally I see results similar to what you reported:
```
[...]

PR #191, opened 2025-08-01
Fix numerical instability in sgarea (boxer)

mcara:
Fixes issues reported in https://github.com/spacetelescope/jwst/issues/9599

### DIFF
--- a/src/cdrizzlebox.c
+++ b/src/cdrizzlebox.c
@@ -88,75 +88,96 @@ This is used by BOXER.
 */
 
 static inline_macro double
-sgarea(const double x1, const double y1, const double x2, const double y2,
-	const int sgn_dx, const double slope, const double inv_slope) {
-  double c, xlo, xhi, ylo, yhi, xtop;
+sgarea(const double x1, const double y1, const double x2, const double y2) {
+    double xlo, xhi, ylo, yhi, xtop;
+    double dx, dy, det, sgn_dx;
 
-  /* Trap vertical line */
-  if (inv_slope == 0) {
-    return 0.0;
-  }
+    dx = x2 - x1;
+    dy = y2 - y1;
 
-  if (sgn_dx < 0) {
-    xlo = x2;
-    xhi = x1;
-  } else {
-    xlo = x1;
-    xhi = x2;
-  }
-
-  /* And determine the bounds ignoring y for now */
-  if (xlo >= 1.0 || xhi <= 0.0) {
-    return 0.0;
-  }
-
-  xlo = MAX(xlo, 0.0);
-  xhi = MIN(xhi, 1.0);
-
-  /* Now look at y */
-  c = y1 - slope * x1;
-  ylo = slope * xlo + c;
-  yhi = slope * xhi + c;
-
-  /* Trap segment entirely below axis */
-  if (ylo <= 0.0 && yhi <= 0.0) {
-    return 0.0;
-  }
-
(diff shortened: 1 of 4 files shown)


mcara:
#### Regression tests:

Romancal: https://github.com/spacetelescope/RegressionTests/actions/runs/16666130761
JWST: https://github.com/spacetelescope/RegressionTests/actions/runs/16666120481

src/cdrizzlebox.c
```diff
@@ -88,75 +88,99 @@ This is used by BOXER.
...
+    if (yhi < 0.0) {
+        yhi = 0.0;
+        xhi = det / dy;
+    }
+
+    if (ylo <= 1.0) {
+        if (yhi <= 1.0) {
+            /* Segment is entirely within the square.
+               The case of zero slope will end up here without ever
+               calling for inv_slope earlier. */
```
t-brandt, line 173: Should adjust the comment to remove the reference to inv_slope

t-brandt (approved):
Looks good to me; thank you.  My only suggested change is to clean up a comment that no longer applies.

braingram:
I rebased https://github.com/spacetelescope/drizzle/pull/187 off of this PR so the CI would run the [test_edge_effects](https://github.com/spacetelescope/drizzle/pull/187/files#diff-3fdee5360ebcd805e85601a588ed6e2279140d4cbc88c21d2c02f38455d74096R1240) unit test added in #187. The unit test passes with the changes in this PR (and the produced wheels also pass when run locally).

I think we can close #187 but I think it's worth keeping the unit test in that PR. @mcara would you copy the unit test over into this PR?

I see a few failures in the jwst regression tests. What's the cause of those differences?

I did not review the code changes in this PR but can do so if it's helpful.

mcara:
> Also since the original motivation for the sgarea changes was to improve performance how does the performance of the code with this PR compare to the original (pre-#175) code?

In the testing that I did, I have noticed a 10% performance degradation compared to the code on `main`. I think this is not too bad considering that previous PR improved performance >2x compared to the previous `compute_area` algorithm, especially that it no longer _seems to_ produce tiny numbers when there is no intersection.

braingram:
> The issue with the #187 is not related to this PR but based on my debugging, it is caused by an issue in _bounding_ polygon intersections. The test from #187 (or from [#189 (comment)](https://github.com/spacetelescope/drizzle/issues/189#issue-3196294879)) will be added to a different PR. BTW, these two seem similar. Which one do you prefer? 

It's up to you. I closed #187 because it seems to be leading to more confusing than helpful.

tapastro:
The JWST regtest results look reasonable to me - the resampled LRS data have a less noisy edge than they used to, but the change is minimal - probably due to how that team defines their specwcs information. I think this is good from JWST side.

braingram (approved):
Thanks! How would you like to coordinate this with jwst? One option would be:
- merge this
- make a new release
- update the jwst pin to use the new release and okify the regtest differences

mcara:
... unless you are OK with merging this, rebasing https://github.com/spacetelescope/drizzle/pull/194 and then comparing regtest failures with (a rebased) https://github.com/spacetelescope/drizzle/pull/194 agains this PR and approve https://github.com/spacetelescope/drizzle/pull/194 if failures are identical. Then we can make a single new release and Okify JWST tests.

mcara:
Re-running regression tests here:

PR #194, opened 2025-08-10
Enhance polygon intersections for nearly collinear edges

### LINKED ISSUE #189: All `nan` output from distorted pixmap

mcara:
Closes #189 

Fixes a bug in the polygon intersection algorithm (used to compute intersections of bounding boxes of input & output images) when one of the edges of one of the polygons is nearly collinear with one of the edges of the other polygon which may result in `NaN` coordinates for the intersection point.


src/cdrizzlemap.c
```diff
@@ -619,12 +666,16 @@ clip_polygon_to_window(const struct polygon *p, const struct polygon *wnd,
...
                 d = area(dp, dw);  // d != 0 because (v2_inside != v1_inside)
+
                 app_ = area(*pv, *pv_);
                 aww_ = area(*wv, *wv_);
+
                 vi.x = (app_ * dw.x - aww_ * dp.x) / d;
                 vi.y = (app_ * dw.y - aww_ * dp.y) / d;
 
-                append_vertex(ppout, vi);
+                if (isfinite(vi.x) && isfinite(vi.y)) {
```
braingram, line 676: Is this the code change that fixed the bug (and all other changes are formatting or unused code)?

src/cdrizzlemap.c
```diff
@@ -442,6 +457,38 @@ is_point_strictly_in_hp(const struct vertex pt, const struct vertex v_,
...
+ * positions. For each rotation, the first vertex is moved to the end of the
+ * vertex array, effectively rotating the polygon's vertex order. The operation
+ * is performed in-place and repeated `pos` times.
+ *
+ * @param p   Pointer to the polygon structure to be rotated. The polygon is
+ *            expected to have an array of vertices `v` and a count `npv`.
+ * @param pos Number of positions to rotate the polygon's vertices.
+ */
+void
+rotate_polygon(struct polygon *p, int pos) {
```
braingram, line 473: Is this used anywhere?

mcara: no, it is not used unless I need to debug stuff: then this comes in handy

src/cdrizzlemap.c
mcara: yes

mcara:
Roman regression tests:

- Using #191 (reference): https://github.com/spacetelescope/RegressionTests/actions/runs/16970032792
- Current PR: https://github.com/spacetelescope/RegressionTests/actions/runs/16970664146

mcara:
JWST regression tests:

- Using #191 (reference): https://github.com/spacetelescope/RegressionTests/actions/runs/16969989506
- Current PR: https://github.com/spacetelescope/RegressionTests/actions/runs/16970674409

PR #203, opened 2025-09-28
Replace 'scale' with two independent parameters: 'iscale' and 'kscale'

mcara:
This PR deprecates parameter `scale` of `Drizzle.add_image()` method. This parameter has the meaning of pixel scale ratio and it was carried over from `drizzlepac` where it is used to do two separate things:
- to rescale input image by `scale**2` because there `PHOTFLAM` of the resampled image -- conversion factor used to convert count rate to flux units -- was not rescaled (and left equal to that of the first input image);
- to rescale resample kernels to the correct size.

These are two different and independent issues and having a single `scale` parameter is just incorrect for a wide range of cases. For example, in JWST, pixel values are surface brightness and converting to flux is achieved by multiplying them by pixel area which is modified for the output image (basically `PHOTFLAM` equivalent is modified) and so a `scale=1` is what is used there. However, this leads to incorrect kernel size, see, e.g., https://github.com/spacetelescope/stcal/pull/412

Splitting `scale` into two independent parameters will allow us to correctly handle various situations/needs.

### DIFF
--- a/drizzle/resample.py
+++ b/drizzle/resample.py
@@ -723,8 +723,9 @@ class Drizzle:
 
         iscale : float, optional
             It is a multiplicative factor used to rescale input image data
-            by ``iscale``. It may make sense to rescale input image by
-            squared pixel scale ratio (the linear dimension of a side of an
+            by ``iscale`` value. ``data2`` images will be rescaled by
+            ``iscale**2``. It may make sense to rescale input image (``data``)
+            by squared pixel scale ratio (the linear dimension of a side of an
             output pixel as seen in the input image's coordinate frame)
             depending on the units of the input image, i.e., counts vs
             brightness. For more details see section
--- a/drizzle/resample.py
+++ b/drizzle/resample.py
@@ -2,6 +2,8 @@
 The `drizzle` module defines the `Drizzle` class, for combining input
 images into a single output image using the drizzle algorithm.
 """
+import warnings
+
 import numpy as np
 
 from drizzle import cdrizzle
@@ -599,9 +601,10 @@ class Drizzle:
 
         return plane_info
 
-    def add_image(self, data, exptime, pixmap, data2=None, dq=None, scale=1.0,
+    def add_image(self, data, exptime, pixmap, data2=None, dq=None,
+                  iscale=1.0, kscale=1.0,
                   weight_map=None, wht_scale=1.0, pixfrac=1.0, in_units='cps',
-                  xmin=None, xmax=None, ymin=None, ymax=None):
+                  xmin=None, xmax=None, ymin=None, ymax=None, **kwargs):
         """
         Resample and add an image to the cumulative output image. Also, update
         output total weight image and context images.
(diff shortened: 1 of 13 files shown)


drizzle/resample.py
```diff
@@ -563,9 +565,10 @@ def _increment_ctx_id(self):
 
         return plane_info
 
-    def add_image(self, data, exptime, pixmap, data2=None, scale=1.0,
+    def add_image(self, data, exptime, pixmap, data2=None,
+                  iscale=1.0, kscale=1.0,
```
braingram, line 569:
```suggestion
                  iscale=1.0, kscale=1.0,
```
Maybe `input_scale_factor` and `kernel_scale_factor`?

perrygreenfield:
As for using pixfrac for kscale, I would argue against that since it would be completely non-obvious that it means something else in a different context. And pixfrac already has a long usage as a name. I would back up keeping these separate parameters and having the software raise an error if both are specified out of context.

perrygreenfield (approved):
LGTM

mcara:
> Mostly I'm trying to understand at the moment if they are in fact mutually exclusive parameters that control related concepts.

Although at first glance these parameters (`kscale` and `pixfrac`) seem to be related -- `pixfrac` controls pixel size of the single pixel in a kernel like "square" and "turbo" while `kscale` controls kernel size for larger (than one pixel) kernels like Lanczos -- I think conceptually they are distinct. For example, in principle, we could get rid of `kscale` parameter because it is used to convert "distances" (or "sizes") from the output image to the input image by using inverse transformation of the `pixmap` - then we would not even need this scale. However, even then, `pixfrac` would still make sense and be used in the square and turbo kernels.

We could also rename `iscale` to `"input_image_pre_resample_multiplicative_factor"` (or drop some of the words, like `pre_factor` but I don't think it would be clearer) and `kscale` - 'pixel_scale_ratio` (that's easier).

mcara:
@braingram @perrygreenfield 

I have updated documentation -- in particular see https://github.com/spacetelescope/drizzle/blob/c5e9095de4e36edc13bdd9910f489e6976dbbdae/drizzle/resample.py#L60-L99 -- and deprecated several parameters in `tblot` that are not really needed and renamed `kscale` as `pixel_scale_ratio`. Please review these lates changes. Thank you!

drizzle/resample.py
```diff
@@ -889,8 +980,9 @@ def add_image(self, data, exptime, pixmap, data2=None, dq=None, scale=1.0,
         return nmiss, nskip
 
 
-def blot_image(data, pixmap, pix_ratio, exptime, output_pixel_shape,
-               interp='poly5', sinscl=1.0):
+def blot_image(data, pixmap, pix_ratio=_DEPRECATED_ARG,
```
braingram: For context, is this used anywhere outside drizzle?

braingram (commented):
Thanks. The docs updates look great and are very clear at explaining when iscale is useful.

Would you run romancal and jwst regtests with this PR?

mcara:
[Romancal regression tests pass](https://github.com/spacetelescope/RegressionTests/actions/runs/18666992170/job/53222048870) if https://github.com/spacetelescope/romancal/pull/2009 and https://github.com/spacetelescope/stcal/pull/418 are merged.

PR #206, opened 2025-10-12
AL-982: Add support for propagation of DQ flags

mcara:
This PR adds support for propagating DQ bitfields from input images to the output image using bitwise-OR.

Also see https://jira.stsci.edu/browse/AL-982

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -21,6 +21,9 @@ Release Notes
   3rd-order kernel. Improved discrete approximation to the Lanczos
   kernel. [#198]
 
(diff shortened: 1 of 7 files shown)


braingram (commented):
A few comments mostly about `int` vs `uint` for the dq arrays.

schlafly (approved):
This looks good to me, thanks Mihai.  Glad that the changes to the C code were pretty self-contained!  We probably need to document that it only works for 32 bit dq masks and 64 bit won't work (which I think is fine).

PR #211, opened 2025-11-20
add pre-commit and apply style fixes

mcara:
Regression tests:
Roman: https://github.com/spacetelescope/RegressionTests/actions/runs/19521543449
JWST: https://github.com/spacetelescope/RegressionTests/actions/runs/19521534471

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,3 +1,3 @@
-# automatically requests pull request reviews for files matching the given pattern; the last match takes precendence
+# automatically requests pull request reviews for files matching the given pattern; the last match takes precedence
 
   *       @spacetelescope/drizzle-maintainers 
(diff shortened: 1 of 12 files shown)


pllim (approved):
Thanks!

But are you sure we should delete CODEOWNERS instead of replacing it with a valid GH team?

mcara:
> But are you sure we should delete CODEOWNERS instead of replacing it with a valid GH team?

If there is such a team - let me know

mcara:
while it may be used by `stcal`, it is a stand-alone data analysis package, isn't it?

mcara:
or maybe `datb-maintainers`?

pllim:
On the other hand, we can keep the file CODEOWNERS but instead of populating it with a team, it would just be `@mcara`.

pllim:
Sure, I am neutral on how this package wants to be managed.

PR #215, opened 2026-01-07
Interpolate WCS coordinates

t-brandt:
This PR implements sparse WCS evaluation and interpolation in lieu of https://github.com/spacetelescope/stcal/pull/367 given the use of drizzle utilities in https://github.com/spacetelescope/stcal/pull/485

It supports bilinear and bicubic interpolation.  Given the NIRCam distortion correction in https://github.com/spacetelescope/jwst-pipeline-notebooks/blob/main/notebooks/NIRCAM/Imaging/JWPipeNB-nircam-imaging.ipynb bicubic interpolation with a stepsize of 10 pixels gives drizzled images identical almost to 32-bit floating point precision.

### DIFF
--- a/drizzle/utils.py
+++ b/drizzle/utils.py
@@ -7,7 +7,8 @@ __all__ = ["calc_pixmap", "decode_context", "estimate_pixel_scale_ratio"]
 _DEG2RAD = math.pi / 180.0
 
 
-def calc_pixmap(wcs_from, wcs_to, shape=None, disable_bbox="to"):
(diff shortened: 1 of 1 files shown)


t-brandt:
Withdrawing in lieu of https://github.com/spacetelescope/stcal/pull/488
The calc_pixmap in that function is now a copy of the calc_pixmap in drizzle with the option to interpolate added, with the goal of enabling a future consolidation of the two routines.

PR #217, opened 2026-01-15
Fix version in the deprecation messages

### LINKED ISSUE #216: Scale deprecation warnings state incorrect version
https://github.com/spacetelescope/drizzle/blob/222ce40d250839c646cb44fb77152e3939f6dc1b/src/cdrizzleapi.c#L304-L310

https://github.com/spacetelescope/drizzle/blob/222ce40d250839c646cb44fb77152e3939f6dc1b/drizzle/resample.py#L815-L820

both mention "version 3.0" (which does not exist). The deprecations were added in 2.2.0.

mcara:
Closes #216 

Corrects version in the deprecation message.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,13 @@ Release Notes
 =============
 
 
+2.2.1 (unreleased)
+==================
+
+- Fixed deprecation warning messages (incorrect version number) introduced in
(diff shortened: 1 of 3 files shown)


braingram (approved):
Thanks! LGTM.

PR #218, opened 2026-01-20
[SCSB-234] use Python 3.13 to build docs with Sphinx 9

zacharyburnett:
Resolves [SCSB-234](https://jira.stsci.edu/browse/SCSB-234)

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -6,9 +6,9 @@ sphinx:
   fail_on_warning: true
 
 build:
-  os: ubuntu-22.04
+  os: ubuntu-24.04
   tools:
-    python: mambaforge-4.10
+    python: mambaforge-latest
 
 conda:
   environment: docs/rtd_environment.yaml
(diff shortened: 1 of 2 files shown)


braingram:
Changes LGTM.

@mcara since you're more familiar with the docs would you look them over to make sure the python version change didn't introduce any unintended changes?

Also FWIW the build did not pick up sphinx 9. I suspect there might be a dependency that is preventing the new version. @zacharyburnett any clue what's limiting the sphinx version?

zacharyburnett:
> Also FWIW the build did not pick up sphinx 9. I suspect there might be a dependency that is preventing the new version. @zacharyburnett any clue what's limiting the sphinx version?

ah, it was because `sphinx_rtd_theme` was included in the conda specs so it was pulling the Conda build of sphinx

braingram:
> > Also FWIW the build did not pick up sphinx 9. I suspect there might be a dependency that is preventing the new version. @zacharyburnett any clue what's limiting the sphinx version?
> 
> ah, it was because `sphinx_rtd_theme` was included in the conda specs so it was pulling the Conda build of sphinx

That makes sense. So for this and other packages we want to pull that out of the conda specs to be able to pick up sphinx 9?

mcara (approved):

PR #219, opened 2026-02-06
Fix tdriz version string to use Python package version

mcara:
Instead of using a hardcoded in C code version string, `tdriz` will read Python package version and report that as C-version string. If retrieving Python package version fails, the code will revert to the hardcoded version so it is still advisable to keep that version string in sync with Python.

Closes [AL-783](https://jira.stsci.edu/browse/AL-783)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,11 +4,10 @@
 Release Notes
 =============
 
-2.2.1 (unreleased)
+3.0.0 (unreleased)
 ==================
 
-- Fixed version string of the C-based drizzle code to be set
-  to the version of the Python package. [#219]
+- Removed version string return value from ``cdrizzle.tdriz()``function. [#219]
 
 
 2.2.1 (unreleased)
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,12 @@
 Release Notes
 =============
 
+2.2.1 (unreleased)
+==================
+
+- Fixed version string of the C-based drizzle code to be set
+  to the version of the Python package. [#219]
+
 
 2.2.1 (unreleased)
 ==================
--- a/drizzle/resample.py
+++ b/drizzle/resample.py
@@ -942,7 +942,7 @@ class Drizzle:
         #       cdrizzleapi.c does not. It should be modified to support this
         #       for performance reasons.
 
-        _vers, nmiss, nskip = cdrizzle.tdriz(
+        nmiss, nskip = cdrizzle.tdriz(
             input=data,
             weights=weight_map,
             pixmap=pixmap,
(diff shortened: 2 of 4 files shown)


braingram:
Is this message even printed anymore? I see:
https://github.com/spacetelescope/drizzle/blob/09bf6902fb7b7acf8a549c1bc9ae6b46b29d49db/drizzle/resample.py#L972

mcara:
@braingram In version 2 I decided to not return this in the Python code. version string was returned pre-2.0.0 - see https://github.com/spacetelescope/drizzle/commit/b850d47cd31139f211eb088fe325cca4f89798e2#diff-18d7445cc4f727ed5801cab7714976aa54a7495a204d38425d5da76f9cdf00e1 (dodrizzle.py).

However, it was left in the C-version of the code - `tdriz` - I think because at that time pipeline(s) and possibly external users may have called `tdriz` directly (because Python code was designed with FITS file handling in mind and was useless for `jwst` and `romancal`).

braingram:
> @braingram In version 2 I decided to not return this in the Python code. version string was returned pre-2.0.0 - see [b850d47#diff-18d7445cc4f727ed5801cab7714976aa54a7495a204d38425d5da76f9cdf00e1](https://github.com/spacetelescope/drizzle/commit/b850d47cd31139f211eb088fe325cca4f89798e2#diff-18d7445cc4f727ed5801cab7714976aa54a7495a204d38425d5da76f9cdf00e1) (dodrizzle.py).
> 
> However, it was left in the C-version of the code - `tdriz` - I think because at that time pipeline(s) and possibly external users may have called `tdriz` directly (because Python code was designed with FITS file handling in mind and was useless for `jwst` and `romancal`).

Thanks. Given that the version string is no longer used what about removing it?

mcara:
> Thanks. Given that the version string is no longer used what about removing it?

It is not used by the Python code but external users calling C code directly expect the function to return 3 values

mcara:
Is there anything wrong with the fix?

braingram:
> It is not used by the Python code but external users calling C code directly expect the function to return 3 values

Is there any known usage of `tdriz`?

> Is there anything wrong with the fix?

If the message is never seen why generate it?

The AL ticket says "While running Roman resample I noticed that drizzle reports a version that is hard coded in the cdrizzleapi module." since this message is no longer displayed we can just close that ticket.

mcara:
> If the message is never seen why generate it?

I search github for possible direct uses of `tdriz()`, and I do not see any (besides some stale forks). I will remove this string.

braingram (approved):
Thanks! Eventually all the python 2 code will be gone :)

PR #224, opened 2026-06-24
Make tblot treat NaN pixmap entries as missing instead of erroring

### LINKED ISSUE #174: Update tblot to handle nan input
The following comment was found in stcal:
https://github.com/spacetelescope/stcal/blob/5f3c27fd6113c5e0c48682900d1dc037d9b8b4b2/src/stcal/outlier_detection/utils.py#L252-L255
>    # Currently tblot cannot handle nans in the pixmap, so we need to give some
>    # other value.  -1 is not optimal and may have side effects.  But this is
>    # what we've been doing up until now, so more investigation is needed
>    # before a change is made.  Preferably, fix tblot in drizzle.

It seems advantageous to update tblot to support nan input values.

jdavies-st:
The C function doblot() used to raise an error whenever the pixmap contained a NaN, forcing the calling function to substitute values (e.g. -1) before calling. NaN comparisons are always false under IEEE-754, so removing the explicit check lets NaN entries fall through to the existing out-of-bounds branch and get filled with `fillval`, matching how the main drizzle path (map_pixel) already treats undefined mappings that are outside the bounds of the target image.

- drop the NaN error branch in doblot() pixmap bounds check, allowing NaNs to be treated as any other mapping that is out-of-bounds
- add test_blot_nan_pixmap unit test

Bonus: convert generator expressions in parametrized tests to `list` types to fix PytestRemovedIn10Warning

Fixes #174

### DIFF
--- a/drizzle/tests/test_resample.py
+++ b/drizzle/tests/test_resample.py
@@ -558,6 +558,25 @@ def test_blot_interpolation(tmpdir, interpolator, test_image_type):
     assert max_diff < 1.0e-5
 
 
+def test_blot_nan_pixmap():
+    """NaN entries in the pixmap (no mapping defined for that output pixel)
+    are treated like an out-of-bounds mapping and filled with ``fillval``
+    instead of raising an error. See issue
+    https://github.com/spacetelescope/drizzle/issues/174
+    """
+    data = np.ones((10, 10), dtype=np.float32)
+    y, x = np.indices((10, 10), dtype=np.float64)
+    pixmap = np.dstack([x, y])
+    pixmap[3, 4] = np.nan
+
+    blotted = resample.blot_image(data, pixmap=pixmap, fillval=42.0)
+
+    assert blotted[3, 4] == 42.0
+    mask = np.ones((10, 10), dtype=bool)
+    mask[3, 4] = False
+    assert np.all(blotted[mask] == 1.0)
+
+
 def test_context_planes():
     """Reproduce error seen in issue #50"""
     shape = (10, 10)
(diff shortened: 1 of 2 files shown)


jdavies-st:
So `stcal.outlier_detection` is the bit of code that does a global replace of NaN with -1 in the pixmaps

https://github.com/spacetelescope/stcal/blob/6cddca3c3d499e0cb541648c5915373cb607eeae/src/stcal/outlier_detection/utils.py#L282-L286

All the unit tests for `stcal` with this new branch of `drizzle` work fine, though I don't think the nan to -1 line is really tested in the unit tests.  But of course we are testing it here in this `drizzle` branch.

```
$ pytest
====================================================== test session starts =======================================================
platform darwin -- Python 3.14.6, pytest-9.1.1, pluggy-1.6.0
rootdir: /Users/jdavies/dev/stcal
configfile: pyproject.toml
testpaths: tests, src/stcal, docs
plugins: cov-7.1.0, doctestplus-1.7.1, ci_watson-0.11.0, requests-mock-1.12.1
collected 724 items                                                                                                              

...

========================================== 724 passed, 13 warnings in 309.99s (0:05:09) ==========================================
```

And similarly the `jwst` package unit tests pass using this `drizzle` branch as well.

```
$ pytest
====================================================== test session starts =======================================================
platform darwin -- Python 3.14.6, pytest-9.1.1, pluggy-1.6.0
rootdir: /Users/jdavies/dev/jwst
configfile: pyproject.toml
plugins: cov-7.1.0, doctestplus-1.7.1, ci_watson-0.11.0, requests-mock-1.12.1
collected 5389 items                                                                                                             

...

=================================== 4003 passed, 1379 skipped, 7 xfailed in 1699.05s (0:28:19) ===================================
```

jdavies-st:
Expanded the tests here and found a pre-existing bug.  See #237.

Wait until #238 (or a similar fix) is merged, rebase, and try again.

PR #227, opened 2026-07-06
[SCSB-237] build Stable ABI3 wheels with `Py_LIMITED_API`

zacharyburnett:
Resolves [SCSB-237](https://jira.stsci.edu/browse/SCSB-237)

building against the Limited API means we can build just ONE wheel (in this case for Python 3.10) and it should work on all future Python versions: https://docs.python.org/3/c-api/stable.html#limited-c-api

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -29,6 +29,7 @@ jobs:
       - uses: actions/setup-python@ece7cb06caefa5fff74198d8649806c4678c61a1 # v6.3.0
         with:
           python-version: "3"
+          pip-install: build
       - if: ${{ runner.os == 'Linux' }}
         uses: docker/setup-qemu-action@96fe6ef7f33517b61c61be40b68a1882f3264fb8 # v4.2.0
         with:
@@ -37,7 +38,6 @@ jobs:
         with:
           package-dir: ./
           output-dir: dist/
-      - run: pip install build
       - run: python -m build --sdist
       - uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
         with:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -19,6 +19,7 @@ jobs:
         runs-on:
           - ubuntu-latest
           - macos-latest
+          - windows-latest
     runs-on: ${{ matrix.runs-on }}
     steps:
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -23,7 +23,9 @@ def get_extensions():
     cfg = {
         "include_dirs": [],
         "libraries": [],
-        "define_macros": [],
+        "define_macros": [
+            ("Py_LIMITED_API", 0x030A0000),  # PY_VERSION_HEX for 3.10
```
zacharyburnett: I set this to Python 3.10 because that's what the current `requires-python` indicates; we can also try 3.9 if we want more backward compatibility for the wheel (though I'm not sure what the differences are with the stable ABI between 3.9 and 3.10). A wheel built for 3.10 stable ABI should work for all future Python versions on or after 3.10

braingram: 3.9 doesn't get security fixes:
https://endoflife.date/python
so let's stick with 3.10.

pyproject.toml
```diff
@@ -48,6 +48,11 @@ requires = [
 ]
 build-backend = "setuptools.build_meta"
 
+[tool.cibuildwheel]
+build = "cp310-*"  # build for CPython 3.10 Stable API
+skip = ["cp3??t-*"]  # explicitly skip free-threaded builds
```
braingram: Let's not stop building 3.14t wheels since we can and are building them on main:
https://github.com/spacetelescope/drizzle/actions/runs/26469716915/job/77939588239#step:9:1299

This does mean this PR is not sufficient to prevent needing a 3.15 release (since we'll need a 3.15t wheel) but hopefully we can use abi3t at that point: https://docs.python.org/3.16/howto/abi3t-migration.html

zacharyburnett:
alright, it now builds a single 3.10 ABI3 wheel, and then also a non-ABI3 wheel for each free-threaded Python version

.github/workflows/build.yml
```diff
@@ -1,28 +1,71 @@
...
-      sdist: true
-      test_command: python -c "from drizzle import cdrizzle"
-    secrets:
-      pypi_token: ${{ secrets.PYPI_PASSWORD_STSCI_MAINTAINER }}
-
+    strategy:
+      matrix:
+        runs-on:
+          - ubuntu-latest
+          - macos-latest
```
braingram, line 21: Would you restore windows wheels?

braingram (changes requested):

zacharyburnett:
running regtest at https://github.com/spacetelescope/RegressionTests/actions/runs/29279579155/job/86917224186

mcara (approved):
LGTM. Thanks!

